PR #41, opened 2016-08-04
update clinvar readme to include instructions for directory structure…

zfisch:
… and running makefile


zfisch:
@melissacline Would love your sign off on this before merging. Hoping to help anybody else along that might need to run this before automation is complete.

PR #122, opened 2016-09-26
Luigi

zfisch:
This includes the Luigi script to stitch all pipeline processes together as well as many updates to the pipeline.

Please note that enigma-processing.py still needs work before it can be used at all -- **it should be determined whether the existing version of enigma-processing.py on master is more correct than the version included in this branch before merging**.


pipeline/data_merging/aggregate_across_columns.py
```diff
@@ -209,27 +213,32 @@ def selectMaxAlleleFrequency(newRow):
                 maxFreq = aa
                 maxFreqString = "%f (AA from ESP)" % aa
     if newRow["EUR_Allele_frequency_1000_Genomes"] != EMPTY:
-        eur_af = float(newRow["EUR_Allele_frequency_1000_Genomes"])
+        freqs = [float(i) for i in newRow["EUR_Allele_frequency_1000_Genomes"].split(',')]
```
zfisch: This allows merging variants with different allele frequencies. Since `newRow["EUR_Allele_frequency_1000_Genomes"]` would be a list in this case, it cannot be converted to a float and compared without treating each item individually.

pipeline/data_merging/variant-merging.py
```diff
@@ -253,35 +281,38 @@ def add_leading_base(chr, pos, ref, alt, version="hg38"):
         brca_pos = pos - 1 - BRCA1[version]["start"]
     else:
         raise Exception("wrong chromosome number")
-    # correct error with when ref is empty string
-    if len(ref) == 0:
-        brca_pos += 1
-    else:
-        assert(seq[brca_pos:brca_pos + len(ref)] == ref)
```
zfisch: These lines created an off by one error for half of the variants. It would probably be a good idea to better understand what error they were intended to fix to make sure it hasn't returned since their removal.

pipeline/enigma/README.md
```diff
@@ -6,39 +6,39 @@ raw_files/ENIGMA_for_BRCAsite_9.21.2015.xlsx
 
 raw_files/ENIGMA_SubmissionClinVar_2016-05-31_ncbi.xlsx
```
zfisch: It looks like this file as well as the enigma-processing.py file are out of sync with master. They could both use review to determine which is more correct.

pipeline/enigma/enigma-processing.py
```diff
@@ -1,93 +1,145 @@
 """
```
zfisch: This file needs review to determine which version is more correct. I believe Lucy is also working on this file.

pipeline/data_merging/aggregate_across_columns.py
melissacline: Yes.  We'll have to address that (and other fields that could have atomic values or could be lists) when we have users wanting to query against these fields...

pipeline/data_merging/aggregate_across_columns.py
```diff
@@ -209,27 +213,32 @@ def selectMaxAlleleFrequency(newRow):
...
+        eur_af = max(freqs)
         if eur_af > maxFreq:
             maxFreq = eur_af
             maxFreqString = "%f (EUR from 1000 Genomes)" % eur_af
     if newRow["AFR_Allele_frequency_1000_Genomes"] != EMPTY:
-        afr_af = float(newRow["AFR_Allele_frequency_1000_Genomes"])
+        freqs = [float(i) for i in newRow["AFR_Allele_frequency_1000_Genomes"].split(',')]
+        afr_af = max(freqs)
```
melissacline: Hmm, this brings up an interesting point.  When multiple variants are merged, should the allele frequency be the max or the sum?  This is a good question for a good computational geneticist....

pipeline/data_merging/variant-merging.py
melissacline: It's amazing how apparent the bug becomes when the relevant lines are highlighted!

PR #144, opened 2016-09-30
resolve error for missing genetic_origin field in lovd data, also upd…

zfisch:
…ate logging and error message


pipeline/data_merging/variant-merging.py
```diff
@@ -556,7 +555,6 @@ def add_new_source(columns, variants, source, source_file, source_dict):
         else:
             variants[genome_coor] = ['-'] * old_column_num
             variants[genome_coor][COLUMN_SOURCE] = source
-            chrm = genome_coor.split(":")[0]
```
zfisch: this variable is never used

PR #161, opened 2016-10-06
update luigi readme to explain arguments

zfisch:


zfisch:
@melissacline here's an updated readme for luigi. Let me know if there is anything else that could use explanation or clarification!

PR #165, opened 2016-10-06
Build metadata script

zfisch:
this is the first pass at a version metadata generation script. note that it requires release notes in the form of a txt file as an argument.

it would be simple enough to integrate this script into the automated pipeline if we want, but for testing I kept it separate for now.


zfisch:
@melissacline I wrote this assuming it would take in a release notes txt file as an argument to generate json to pass along with the built.tsv file. all of the other version data is generated downstream (links to files and md5sum etc.).

zfisch:
@maryjgoldman is there a template release notes file you'd like us to use for versioning?

maryjgoldman:
I don't have a template. I thought I saw a .json file in the new new data? Can we just edit that? It really is whatever file type/template would work best for slurping into Django

PR #167, opened 2016-10-07
Add change types to built

zfisch:
releaseDiff now outputs an updated built_with_change_types.tsv file with appropriate change_types for each variant.


pipeline/utilities/releaseDiff.py
```diff
@@ -201,6 +220,33 @@ class v1ToV2(transformer):
         }
 
 
+def appendVariantChangeTypesToOutput(variantChangeTypes, v2, output):
```
melissacline: Nice - this makes it so simple!

PR #186, opened 2016-10-12
Restructure output dir

zfisch:
1. Organize output directory for easier navigation and storage.
2. Store log files in output/artifacts/
3. Helps with compressing the directory and describing the different files and locations.


zfisch:
If it would be helpful for debugging or if it's relevant to people who want to download the data, I think this would make sense. Otherwise it may just be clutter in the output directory.

melissacline:
It might be helpful for debugging.

zfisch:
sounds good, let me know if you'd like me to implement that as part of this PR or for a future PR.

melissacline:
Let's go ahead and implement it.  Since almost no one will download this tarball besides us, it won't affect (almost) anyone else, but it could help us.

zfisch:
on it, will update this PR when it's ready.

zfisch:
@melissacline just updated this PR to replace the temp directory with the artifacts directory in variant-merging.py, double checking for re-approval before merging.

PR #216, opened 2016-10-28
Remove ga4gh dependency

david4096:
Moves the python notebook to the static folder and removes `ga4gh` as a requirement in favor of the lighter `ga4gh-schemas` @zfisch


.gitignore
```diff
@@ -1,2 +1,3 @@
 *.DS_Store
 pipeline/brca/*
+.ipynb_checkpoints
```
zfisch: needs a newline

zfisch:
@david4096 looks like circleci failed for some reason -- sometimes rerunning it solves a problem if everything looks clear.

david4096:
It seems like it failed due to something related to the release mechanics. There don't seem to be any tests of the django code directly.

PR #218, opened 2016-10-31
add protobuf to requirements.txt

zfisch:
@david4096 I got an error regarding a missing requirement for protobuf and added it to requirements.txt. I'll await your approval on this one -- not sure if this is the best version.


david4096:
Looks good to me, might be what was wrong with my other PR.

PR #241, opened 2016-11-08
Release note update239

melissacline:
Updated the release notes for both releases


zfisch:
@melissacline I can derive the needed information from this PR, but we should close it and not merge. When updating release notes, md5s and release archives need updating as well. In this case, the version.json file for 10-21 is no longer relevant (will be altogether replaced by a 11-5 version.json file).

In the future, the easiest way to update release notes is to send new notes over in a txt file to be referenced in a couple final pipeline scripts. If it's needed we can write a script for updating release notes in the future.

PR #258, opened 2016-11-24
Variant history

strbean:
Show the history of changes to  a variant in the variant detail page.

More feedback needed before merge.


website/js/index.js
```diff
@@ -448,6 +479,84 @@ var VariantDetail = React.createClass({
...
+            "Clinical_Significance_ClinVar",
+            "Allele_Origin_ClinVar"
+        ];
+        for (var i = 0; i < data.length; i++) {
+            let version = data[i],
+                changes = [],
+                release = version["Data_Release"], //eslint-disable-line dot-notation
+                hightlightRow = false;
+
+            // if this is not the oldest version of the variant, diff them
```
zfisch: im a little confused about the intention here, could you clarify?

im assuming it's supposed to diff the current version of a variant with whatever the previous version was and nothing more.

website/js/index.js
```diff
@@ -448,6 +479,84 @@ var VariantDetail = React.createClass({
...
+                            if (added.length || deleted.length) {
+                                changes.push(
+                                    <span>
+                                        <strong>{ getDisplayName(key) }: </strong> <br />
+                                        { added.join(', ') }{ !!(added.length && deleted.length) && ', '}{ deleted.join(', ') }
+                                    </span>, <br />
+                                );
+                            }
+                        }
+                        else if (_.contains(listKeys, key)) {
```
zfisch: it looks like this case is almost identical to the pathogenicity case -- can they be combined and encapsulated into a dryer function call?

website/js/index.js
zfisch: nevermind -- i can see that we're displaying diffs between every version.

website/js/index.js
```diff
@@ -466,6 +575,26 @@ var VariantDetail = React.createClass({
                 </Row>
                 <Row>
                     <Col md={8} mdOffset={2}>
+                        { /* eslint-disable dot-notation */ }
+                        <h3>{variant["HGVS_cDNA"]}</h3>
```
zfisch: I see these everywhere: `/* eslint-disable dot-notation */`

Is there a reason to write it as `variant["HGVS_cDNA"]` and disable the linter rather than `variant.HGVS_cDNA` as the linter wants?

website/js/util.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 var months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"];
-var dateFormat = (date) => { let d = new Date(date); return `${months[d.getMonth()]} ${d.getDate()}, ${d.getFullYear()}`; };
```
zfisch: Changing this will affect the way the data is displayed on the previous releases page (see http://brcaexchange.org/releases, would turn e.g. `October 6, 2016` into `6 October 2016`). Is this intentional? 

We could also use a library like Moment.js or Date.js or any other number of other libraries to handle this for us along with other date formatting issues we might hit -- common js problem.

website/js/index.js
strbean: The linter complains about that as well, for not being camel case.

website/js/util.js
strbean: I figured date display should be consistent.

I took a look at those options and couldn't decide so I opted to do this. I'd be all for moving to a library though.

website/js/index.js
strbean: The case for pathogenicity_all is pretty weird. This could definitely use some refactoring, I think it might be best to move the variant history table into it's own component even.

strbean: I can start in on this next week. It's finals week but almost all my finals are on Monday, I should be able to get it done on Tuesday.

strbean:
Adopted moment.js, and did some refactoring on the list-diffs.

website/js/index.js
strbean: Yeah, I think disabling dot-notation would be good.

zfisch: Please make this change and remove all references to disabling the dot-notation in eslint.

zfisch: Will be done in a followup

PR #265, opened 2016-11-28
Deleted Variants

strbean:
For comparision, not ready for merge.


website/js/DataTable.js
```diff
@@ -23,10 +23,11 @@ var merge = (...args) => _.extend({}, ...args);
 
 var Lollipop = require('./d3Lollipop');
 
-function setPages({data, count, synonyms}, pageLength) {
+function setPages({data, count, deleted_count, synonyms}, pageLength) { //eslint-disable-line camelcase
```
zfisch: same case as before with eslint -- should decide whether to follow the rule or remove it.

website/django/data/views.py
```diff
@@ -133,7 +140,7 @@ def index(request):
         count = query.count()
         query = select_page(query, page_size, page_num)
         # call list() now to evaluate the query
-        response = JsonResponse({'count': count, 'synonyms': synonyms_count, 'data': list(query.values(*column))})
+        response = JsonResponse({'count': count, 'deleted_count': deleted_count, 'synonyms': synonyms_count, 'data': list(query.values(*column))})
```
zfisch: if we want to follow the camelcase rule, it could be fixed here by making `'deleted_count' == 'deletedCount'`

website/django/data/views.py
```diff
@@ -105,6 +108,10 @@ def index(request):
     if search_term:
         query, synonyms_count = apply_search(query, search_term, quotes=quotes, release=release)
 
+    if not show_deleted and not release:
```
zfisch: do we need to support cases where users want to see deleted variants for a specific release?

strbean: Currently, they are viewable like other change types, but first thing you see is a blank table and the "N deleted variants match" text. I mentioned this to Melissa and Mary and they said it was low priority, but it should be a quick fix.

website/django/data/views.py
strbean: Good call.

website/django/data/views.py
strbean: Ah, scratch that, viewing deleted variants for a release is working fine.

zfisch: 👍

PR #279, opened 2016-12-09
run eslint as part of webpack

zfisch:
Background: We currently use es-lint to ensure code consistency, but it has to be run manually using `npm run lint` otherwise errors will not be seen until an attempted PR is rejected (es-lint is run by circle-ci).

This change would automatically run ES-Lint as part of `npm run` through webpack so errors would be brought up immediately.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -24,6 +24,7 @@
     "css-loader": "^0.9.1",
     "d3-dsv": "^0.1.5",
     "eslint": "^1.10.3",
+    "eslint-loader": "^1.6.1",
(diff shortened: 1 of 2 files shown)


zfisch:
@strbean Thoughts? I don't want to merge this if it will impede your workflow. I never remember to `npm run lint` so I figured I'd have webpack do it for me, but I can start remembering if I need to.

strbean:
This would be awesome!

Brian mentioned something about doing this as part of a git commit hook, but that never happened and I don't know anything about implementing it, seems like this is just as good!

PR #302, opened 2016-12-14
disable the dot-notation rule in .eslintrc

strbean:
Disable the dot-notation rule, and remove per-line 'eslint-disable' comments.

Also, switch to dot notation in a few instances where it doesn't cause any problems.


acthp:
Not sure you want to disable that eslint rule. What was the problem with it? It flags inconsistent style, and cross-browser problems (reserved words). It can raise false positives for objects used as dictionaries (rather than records), if you reference a dictionary key as a literal.

I think the exceptions in the code were to work around that case, like data['email_confirm']. The style emphasizes that this is a dictionary (variable keys) not a record (fixed keys), but the rule flags it for being constant.

Another way to work around it (vs. an eslint directive) is to put the keys in a var somewhere, like 

var EMAIL_CONFIRM = 'email_confirm'

or

var fields = {
  email_confirm: 'email_confirm',
  ...
};

I don't have strong feelings about it, but generally like stronger linting rules to make the code easier to read.

strbean:
Another possibility would be using the allowPattern option to exempt 'snake-case' from the rule. This would cover all the field names from the variant table.

strbean:
Re-enabled the rule and added allowPattern to allow snake_case names.

PR #318, opened 2016-12-21
Filter exLOVD citation format change out of change log. Closes #294

### LINKED ISSUE #294: Filter out changes to literature references when the first author is the same.
See for example:  Literature Reference (exLOVD): Easton DF et al., Am J Hum Genet, 81: 873-883, 2007. Easton et al. 2007 
http://brcaexchange-dev.cloudapp.net:9001/variant/22376

strbean:


website/js/index.js
```diff
@@ -537,10 +537,16 @@ var VariantDetail = React.createClass({
...
                             }
                         } else {
+                            // If date changed from YY format to YYYY format, ignore. Ex: 1/10/15 -> 1/10/2015
                             if (key === "Date_last_evaluated_ENIGMA" &&
                                 moment(version[key], "MM/DD/YYYY").format("DD MMMM YYYY") === moment(version[key], "MM/DD/YYYY").format("DD MMMM YYYY")) {
                                 continue;
                             }
+                            // exLOVD citation format changed, heuristic for matching: first word (i.e. first author) same -> ignore
+                            else if (key === "Literature_source_exLOVD" &&
+                                version[key].trim().split(' ')[0] === previous[key].trim().split(' ')[0]) {
```
strbean: Yeah, that's what Melissa and Mary decided on.

maryjgoldman:
Yes, I designed this. There is a chance that the citation could change and the first author would be the same, but this is unlikely. And if it does happen the change is more likely to be correction of a citation rather than a new citation all together.

PR #340, opened 2016-12-27
Closes #327 -- ignore cases where order changes during diff process

### LINKED ISSUE #327: In releaseDiff.py, ignore delimited fields in which the same elements are in a different order
Here is an example of the issue:

 chr13:g.32341140:T>G 
 Old Source: ClinVar,BIC 
 New Source: ClinVar,BIC 

Pathogenicity_all: Uncertain_significance,Likely_benign (ClinVar); Pending (BIC) | Likely_benign,Uncertain_significance (ClinVar); Pending (BIC) 

There's already something in the code to do an order-independent comparison of delimited lists, but  that code probably needs to be generalized for this case, where the list has two different delimiters.

zfisch:
It turns out we only delimit using a semicolon on the `Pathogenicity_all` field, which means the comma delimited search we already have in place should catch other reorders.

Given that, I found that simply checking to make sure the exact same characters are in both the concatenated new and old values for `Pathogenicity_all` is a reliable check.

### DIFF
--- a/pipeline/utilities/releaseDiff.py
+++ b/pipeline/utilities/releaseDiff.py
@@ -69,7 +69,7 @@ class transformer(object):
         listsAreConsistent = False
         if oldValues is None or newValues is None:
             return False
-        elif field == "Pathogenicity_all":
+        elif re.search(";", oldValues) and re.search(";", newValues):
             return equivalentSemicolonDelimitedValues(oldValues, newValues)
         elif re.search(",", oldValues) and re.search(",", newValues):
             oldTokens = oldValues.split(",")
--- a/pipeline/utilities/releaseDiff.py
+++ b/pipeline/utilities/releaseDiff.py
@@ -69,8 +69,8 @@ class transformer(object):
         listsAreConsistent = False
         if oldValues is None or newValues is None:
             return False
-        elif re.search(";", oldValues) and re.search(";", newValues):
-            return equivalentSemicolonDelimitedValues(oldValues, newValues)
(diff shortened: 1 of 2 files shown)


zfisch:
Just realized a potential issue if classifications swapped between sources (e.g. `Benign (ClinVar); Pathogenic (BIC)` -> `Pathogenic (ClinVar); Benign (BIC)`) would not be caught. Update coming soon!

zfisch:
Updated to address the aforementioned issue and add unit tests.

melissacline:
> add unit tests
w00t!!!

zfisch:
@melissacline unfortunately the issue is slightly more complex than splitting on semicolons and then splitting on commas (e.g. if we split `Uncertain_significance,Likely_benign,not_provided (ClinVar); Pending (BIC) | not_provided,Likely_benign,Uncertain_significance (ClinVar); Pending (BIC)` on semicolons and then commas, it would return a false negative of `not_provided (ClinVar) != Uncertain_significance (ClinVar)` even though the original values are in fact equivalent. This is the reason I switched to a character comparison instead.

I did make an update, however, to use the sorted comparison in this PR for all semicolon delimited values instead of only `Pathogenicity_all`. That being said, only `Pathogenicity_all` and `Condition_ID_value_ENIGMA` have semicolons.

zfisch:
Ack, it actually requires special treatment with or without semicolons. As such, I renamed the function to clarify that it is an edge case particular to `Pathogenicity_all`.

PR #346, opened 2017-01-04
Script to overwrite local db with beta or prod

zfisch:


zfisch:
@strbean let me know if there are any other options that would be useful to you in this script.

PR #368, opened 2017-01-23
Closes #367

### LINKED ISSUE #367: signup for mailing list is broken
Uncaught SyntaxError: Unexpected token < in JSON at position 1
    at JSON.parse (<anonymous>)
    at XMLHttpRequest.xhr.onload (MailingList.js:194)

zfisch:
needs tests before merge


zfisch:
@strbean there's a thousand ways to approach this, but let me know if you can think of some particularly userful tests around the mailchimp API / mailing list signup

strbean:
The simplest thing I think would be to just grab the /lists endpoint and see ensure the response code is OK. That would catch any config problems, although I guess that could cause a build to fail because of external errors (if MailChimp goes down or something). Not sure what the usual approach is to this.

zfisch:
Yea, I think this is ok for now. If we run into further issues with it, we can add tests to ensure whatever bugs we patch stay patched.

PR #377, opened 2017-01-26
Load diff json

strbean:
Ready for review


website/django/data/management/commands/add_diff_json.py
```diff
@@ -0,0 +1,25 @@
+from django.core.management.base import BaseCommand, CommandError
+# from django.conf import settings
```
zfisch: pls remove if not needed

strbean: done

PR #383, opened 2017-01-31
Fix release diff miss equal proteins

zfisch:
This PR addresses false diffs generated by diffing deprecated fields from built.tsv. For context:

zfisch [6:34 PM] 
^ we were previously including changes to the non-pyhgvs columns in our diff calculations. this change ensures our diff.json files do not reflect those false diffs -- though it does mean some of our change_types for previous releases are incorrect

zfisch [6:41 PM] 
e.g. variant `chr17:g.43091495:C>T` is currently considered changed in release 3 because it thinks the HGVS_Protein value changed from `NM_007294:p.Val1809Phe` to `p.(Val1809Phe)`. In reality, this is the deprecated HGVS_Protein value from built.tsv that changed. The more significant value, `pyhgvs_Protein`, is the same for all versions of the variant -- `NP_009225.1:p.(Val1809Phe)`. This is also the only value mentioned that gets stored in the db (confusingly, it's stored as `HGVS_Protein` even though it comes from the `pyhgvs_Protein` column). (edited)

### DIFF
--- a/pipeline/utilities/releaseDiff.py
+++ b/pipeline/utilities/releaseDiff.py
@@ -159,7 +159,8 @@ class transformer(object):
 
         # Uncomment if using old data schema (e.g. pre pyhgvs_Genomic_Coordinate_38)
         columns_to_ignore = ["change_type", "Assertion_method_citation_ENIGMA", "Genomic_Coordinate_hg36",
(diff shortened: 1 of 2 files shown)


zfisch:
Working on some further testing on this branch -- pls wait to merge until further improvements have been pushed up.

zfisch:
@strbean this is now ready for review. there's lot more we could do for testing here, but this is a solid start. I'd recommend running this over the dec/nov releases to produce a diff to test in the UI.

PR #387, opened 2017-01-31
Use materialized view for GA4GH API

### LINKED ISSUE #386: GA4GH API not returning all variants
As observed by @almussel the GA4GH API is serving some subset of the variants hosted by the BRCA exchange:
```
from ga4gh.client import client
c = client.HttpClient("http://brcaexchange.org/backend/data/ga4gh/v0.6.0a7") 
# This takes a few seconds
brca1 = len(list(c.search_variants(variant_set_id='brca-hg37', start=0, end=123123123123, reference_name="17")))
# 2242
brca2 = len(list(c.search_variants(variant_set_id='brca-hg37', start=0, end=123123123123, reference_name="13")))
# 2824
```
I suspect this might be due to how the data releases are being handled and that only differences are added to new releases. https://github.com/BD2KGenomics/brca-exchange/blob/master/website/django/data/views.py#L355

Is there something else that could be causing us to be getting so few variants? What is the correct way to get all of the variants for these variant sets? Once we've figured out a way to present only the latest, but not just the differences, we can evaluate added variant sets that go with different data releases.

Thanks!

david4096:
The GA4GH API was incorrectly using the database view. As suggested by @strbean replaces the `latest_variants` function with the CurrentVariant view.

Also replaces two `BadRequest` calls with the proper HttpResponse.  close #386

### DIFF
--- a/website/django/data/views.py
+++ b/website/django/data/views.py
@@ -349,11 +349,6 @@ def sanitise_term(term):
     term += ":*"
     return term
 
-def latest_variants():
-    """Helper function provides variants from the latest release"""
-    latest = DataRelease.objects.order_by('-id')[0].id
-    variants = Variant.objects.filter(Data_Release_id = latest)
-    return variants
 
 @require_http_methods(["POST"])
 def search_variants(request):
@@ -379,7 +374,7 @@ def search_variants(request):
         page_token = '0'
 
     response = variant_service.SearchVariantsResponse()
-    variants = latest_variants()
+    variants = CurrentVariant.objects
     dataset_id, reference_genome = variant_set_id.split('-')
     if dataset_id != DATASET_ID:
         return HttpResponseBadRequest(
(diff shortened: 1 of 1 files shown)


zfisch:
Looks good!

BTW, we're now running django tests as part of the circleCI builds using pytest. If you'd like to include the tests for the GA4GH api, you could change `tests.py` to be something like `test_GA4GH.py`, and make a few updates to the import statements and setup (similar to those in `brca-exchange/website/users/test_views.py`, requirements need updating in `brca-exchange/test-requirements.txt`). The command that runs the tests is in the `brca-exchange/circle.yml` file: `cd .. && pip install -r test-requirements.txt && pytest --junitxml=${CIRCLE_TEST_REPORTS}/pytest-results.xml`

almussel:
Is there something holding this up? I'd love to be able to take advantage of these changes ASAP.

PR #412, opened 2017-02-23
Fix cosmetic diffs

zfisch:
This PR accomplishes a few things:

1. `_makeExpectedChanges` was written to handle the differences between release 1 and 2. It's guts have been refactored into the `normalize` method for generalization.
2. Lots of updates have been made to catch non-changes, particularly regarding pathogenicity updates. For example, it's possible that the data has duplicate entries for pathogenicity in one release and only one in another -- in this case, "Pathogenic, Pathogenic (ClinVar)" and "Pathogenic (ClinVar)" can be considered a non-change (though it should be resolved earlier in the pipeline eventually to prevent these duplicates).
3. Reorders are handled more thoroughly by using the LIST_KEYS list rather than simply checking for commas in the values.

The result is a reduction in change_types by 3-4X.

### DIFF
--- a/pipeline/utilities/releaseDiff.py
+++ b/pipeline/utilities/releaseDiff.py
@@ -130,28 +130,42 @@ class transformer(object):
             value = value[:len(value)-1]
 
         if field == "Submitter_ClinVar":
+            # Nagging trailing underscore and other random disparities...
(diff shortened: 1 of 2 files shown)


zfisch:
Much Better:

change_types_before_fix
change_types_after_fix

PR #446, opened 2017-04-13
All Data Portal UI Redesign

falquaddoomi:
This PR introduces a few large changes to the All Data Portal Variant Details page, specifically:

1. Fields are now displayed in groups, the structure of which is defined in the 'groups' variable in js/VariantTable.js.
2. Depending on the width of the viewport, the groups reflow into a maximum of three columns and a minimum of one column. Isotope (and specifically the Packery layout engine) is used to minimize the vertical gaps between elements.
3. The visibility of empty fields (and groups that contain exclusively empty fields) is toggle-able from a notification box at the top of the page. This setting is retained in local storage and applied on subsequent views.
4. Clicking the header of a group toggles between expanding and collapsing its associated fields. The expanded state of each group is retained in local storage and applied on subsequent views.
5. Fields that contain commas have their commas replaced with comma-spaces to allow them to break to a new line if necessary. Empty elements (either blank or a single hyphen) in comma-delimited lists are removed. Fields that are blank are replaced with hyphens.
6. To retain the appearance of the Expert-Reviewed portal, the VariantDetail component now contains a dispatch in render() that, if the page is in "expert-reviewed mode", calls the older version of the render() function  (renamed to renderExpertReviewed()) that renders a flat table of fields.

Due to an interaction between the version of react-bootstrap that we're using (0.20.8) and Isotope/Packery, expanding a group causes it to temporarily overlap the elements that follow it. This is rectified once the header has fully expanded by invoking Packery to re-layout the page.


website/js/index.js
```diff
@@ -588,15 +705,196 @@ var VariantDetail = React.createClass({
...
+                        rowItem = <i>{variant[prop]}</i>;
+                    }
+                    else if (prop === "URL_ENIGMA") {
+                        if (variant[prop].length) {
+                            rowItem = <a target="_blank" href={variant[prop]}>link to multifactorial analysis</a>;
+                        }
+                    } else if (prop === "Assertion_method_citation_ENIGMA") {
+                        rowItem = <a target="_blank" href="https://enigmaconsortium.org/library/general-documents/">Enigma Rules version Mar 26, 2015</a>;
+
+                        // this will be used in All Data display
```
zfisch: can rem?

website/js/index.js
```diff
@@ -588,15 +705,196 @@ var VariantDetail = React.createClass({
...
+                    } else if (prop === "HGVS_cDNA") {
+                        rowItem = variant[prop].split(":")[1];
+                    } else if (prop === "HGVS_Protein") {
+                        rowItem = variant[prop].split(":")[1];
+                    } else if (prop === "Date_last_evaluated_ENIGMA" && !isEmptyField(variant[prop])) {
+                        rowItem = moment(variant[prop], "MM/DD/YYYY").format("DD MMMM YYYY");
+                    } else {
+                        // FIXME: make this less of a hack in the future
+                        // make sure commas wrap
+                        if (variant[prop]) {
```
zfisch: can you wrap all of this normalization logic into a function for easier use/management?

website/js/css/custom.css
```diff
@@ -702,12 +703,41 @@ div.variant-detail-group .panel {
...
+.Variant-detail-headerbar {
+    text-align: right;
+    vertical-align: bottom;
+}
+
+.vcenterblock {
+    display: inline-block;
+    vertical-align: middle;
+    float: none;
 }
```
zfisch: needs new line at end

website/js/index.js
```diff
@@ -434,7 +430,49 @@ function isEmptyDiff(value) {
     return value === null || value.length < 1;
 }
 
-// let masonryHack = null;
+// attempts to parse the given date string using a variety of formats,
+// returning the formatted result as something like '08 September 2016'.
+// just returns the input if every pattern fails to match
+function normalizeDateFieldDisplay(value) {
```
zfisch: is any of this necessary for date formatting? I know we use moment.js (search for the use of moment in js/index.js and js/Releases.js) to handle date formatting elsewhere -- it seems we should be able to make use of this library throughout the app rather than writing custom functions. Also, shout out to how frustrating date formatting is in general.

website/js/css/custom.css
```diff
@@ -642,3 +643,101 @@ th[role='columnheader'] .help, .help-target .help {
...
+.Variant-detail-headerbar {
+    text-align: right;
+    vertical-align: bottom;
+}
+
+.vcenterblock {
+    display: inline-block;
+    vertical-align: middle;
+    float: none;
+}
```
zfisch: needs new line

website/js/index.js
```diff
@@ -416,12 +430,99 @@ function isEmptyDiff(value) {
...
 
+// attempts to parse the given date string using a variety of formats,
+// returning the formatted result as something like '08 September 2016'.
+// just returns the input if every pattern fails to match
+function normalizeDateFieldDisplay(value) {
+    // extend this if there are more formats in the future
+    const formats = ["MM/DD/YYYY", "YYYY-MM-DD"];
+
+    for (let i = 0; i < formats.length; i++) {
+        const q = moment(value, formats[i]);
```
zfisch: should this be `let` instead of `const`?

falquaddoomi: Considering that 'q' is never reassigned, I think it's fine to leave it const? Const doesn't indicate anything about the immutability of the object itself, just that the association between the symbol 'q' and that value can't be changed once assigned.

website/js/css/custom.css
falquaddoomi: Good catch!

website/js/index.js
zfisch: checking back in on this, is this block necessary or can we remove it?

falquaddoomi: apologies, forgot to respond to this one. i didn't write that code, so i was reluctant to remove it, but sure, i don't think it's doing anything commented-out in there aside from making block even more difficult to read. i'll go ahead and remove it and otherwise fix up the formatting there.

website/js/css/custom.css
zfisch: it's actually complaining about needing a new line at the end of the file. this is because when it concatenates and minifies files it can cause errors if there's no lines between them.

website/js/index.js
zfisch: 👍  big fan of removing dead code.

PR #454, opened 2017-04-21
Fact sheet

strbean:
Adds Facts and Stats page with lots of static content and live totals for variants.

Also remove old links from navbar, remove CIMB logo from front page, and add ETH Zurich and UCSC Genomics Institute logos.


website/js/FactSheet.js
```diff
@@ -0,0 +1,152 @@
...
+var {Link} = require('react-router');
+
+var FactSheet = React.createClass({
+    mixins: [PureRenderMixin],
+    getInitialState: function() {
+        return {};
+    },
+    componentWillMount: function() {
+        backend.variantCounts().subscribe(
+            resp => this.setState(this.setState(resp)),
```
zfisch: never seen this before (setState inside a setState call), is this intentional and necessary?

website/js/FactSheet.js
```diff
@@ -0,0 +1,152 @@
...
+            () => this.setState({error: 'Problem connecting to server'}));
+    },
+    render: function () {
+        //var {data, page, totalPages, error} = this.state;
```
zfisch: pls remove comment if not needed

strbean:
I'll fix all those issues right away! No idea how I managed to put that
nested setState, I'm surprised it didn't blow up when I tested it ><

strbean:
Alright, I fixed those issues. I removed images that were made redundant by logo changes, but I didn't do a thorough search to see if any of the older images were already unused.

PR #486, opened 2017-05-05
store filter and column selections in local storage, fix table layout…

zfisch:
This PR changes a few things and requires further review before merge.

1. Introduces local storage to manage column, filter, and source selections for the variant table.
2. Ignores query string when determining column, filter and source selections.
3. Fixes a table layout bug when adding new columns to the variant table.

Outstanding questions:

1. Should local storage override query parameters?
2. Should filters be persistent in the same way column selections are persistent? Furthermore, should they persist between portals?

### DIFF
--- a/website/js/DataTable.js
+++ b/website/js/DataTable.js
@@ -101,11 +101,15 @@ var DataTable = React.createClass({
         this.fetch(this.state);
     },
     getInitialState: function () {
+        let filterValues = JSON.parse(localStorage.getItem('filterValues'));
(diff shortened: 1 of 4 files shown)


zfisch:
@strbean @falquaddoomi I've included you as reviewers on this PR since it introduces some relevant changes -- I'm currently awaiting further instruction from @melissacline about how we want these changes implemented and will ping you for discussion afterward.

zfisch:
I'm closing this PR to make some changes as requested in our last development meeting. A new one will be submitted when the changes are ready for review.

Issue #478, opened 2017-05-04
Community page cleanups


melissacline:
Two special UI cleanups have been requested:
1. Enlarge the map, perhaps to scale with the width of the content below.  In the past, we'd had specific feedback to make the map smaller, but now that more members have signed up, we arguably need to make it bigger for better resolution.
2. Somewhere, indicate the number of members who have signed up.  Perhaps, next to the controls, show some text such as "1-10 out of 50,000"

falquaddoomi:
Should the map retain its aspect ratio (i.e. grow vertically) as it scales horizontally? Giving it a fixed (perhaps slightly larger) vertical size and having it stretch only horizontally is somewhat easier to implement.

PR #490, opened 2017-05-08
Community page cleanups #478

falquaddoomi:
Community page map now scales horizontally to match the size of the user-list column and is slightly taller than before (300px vs. 250px.) Added an indicator of the currently-visible and total number of members on the bottom-left of the user-list.


melissacline:
@strbean Would you do a quick review of this PR?

PR #503, opened 2017-05-15
Save column filter selections

zfisch:
This PR introduces local storage for managing column and filter selections and resolves several issues pertaining to table layout, local storage vs. query parameters vs. defaults, and bugs in expert mode. Changes are outlined below:

1. All filter selections and column selections are stored in local storage.
2. Several bugs regarding management of column selections and filter selections are now resolved such that query parameters correctly reflect applied selections.
3. Query params override local storage which overrides default settings. If neither query params nor local storage selections are available, default settings are used.
4. Fixes a bug so that expert mode always shows all columns and all sources.
5. Fixes a table layout issue when adding columns.
6. Fixes a bug when filtering for `Not Yet Reviewed` variants.

### DIFF
--- a/website/js/DataTable.js
+++ b/website/js/DataTable.js
@@ -101,11 +101,15 @@ var DataTable = React.createClass({
         this.fetch(this.state);
     },
     getInitialState: function () {
+        let filterValues = JSON.parse(localStorage.getItem('filterValues'));
+        if (filterValues === null || filterValues === undefined) {
+            filterValues = {};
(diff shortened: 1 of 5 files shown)


zfisch:
Further additions to this PR now include:

- `Restore Defaults` button added to clear all filter/source/column/change_type/release selections are restore default settings.

- `Clear Filters` button previously available when specific releases or change_types are being viewed is no longer available.

PR #586, opened 2017-08-01
Fands feedback

zfisch:


zfisch:
@maryjgoldman @strbean I made the last change, merging now

PR #597, opened 2017-08-11
Allele frequency chart

strbean:
Still needs review / feedback.


strbean:
Ignore, branch name typo ><

website/js/AlleleFrequencyCharts.js
```diff
@@ -0,0 +1,158 @@
+'use strict';
+
+var React = require('react'),
+    BarChart = require('./BarChart'),
+    _ = require('lodash');
+
+// copied from index.js, move to utility file
```
zfisch: if it's straightforward to move to a utility file, may as well do it now before we have duplicate code

website/js/VariantTable.js
```diff
@@ -122,6 +123,17 @@ const researchModeGroups = [
     ]},
 
     {groupTitle: 'Allele Frequency Reference Sets', internalGroupName: 'Allele Frequency Reference Sets', innerCols: [
+        {title: 'Maximum Allele Frequency (1000 Genomes, ESP, ExAC minus TCGA)', prop: 'Max_Allele_Frequency', core: true},
```
zfisch: Max Allele Frequency was recently removed, it should be removed here as well

PR #599, opened 2017-08-12
implement tests for general search queries

collinabidi:


website/django/data/searchtest.py
```diff
@@ -0,0 +1,947 @@
...
+
+            self.assertIsInstance(response, JsonResponse, message)
+            self.assertEqual(response.status_code, 200, message)
+
+            response_data = json.loads(response.content)
+
+            self.assertEqual(response_data['count'], 1, message)
+
+            response_variant = response_data['data'][0]
+            self.assertEqual(response_variant[field_2], getattr(self.existing_variant,field_2), message)
```
zfisch: might as well also add `self.assertEqual(response_variant[field_1], getattr(self.existing_variant, field_1), message)`

website/django/data/searchtest.py
```diff
@@ -0,0 +1,947 @@
...
+
+        self.assertEqual(response_data['count'], 1)
+
+        response_variant = response_data['data'][0]
+        self.assertEqual(response_variant['HGVS_Protein'], self.existing_variant.HGVS_Protein)
+
+
+    def test_search_by_hgvs_protein_and_protein_change_with_colon(self):
+        '''Tests searching for 'HGVS_Protein:Protein_Change' with colon separator
+        NP_009225.1:A280G --> HGVS_Protein.split(':')[0] Protein_Change'''
```
zfisch: this should be: `NP_009225.1:A280G --> HGVS_Protein.split(':')[0]:Protein_Change'''` (notice the added colon between hgvs protein and protein change)

website/django/data/searchtest.py
```diff
@@ -211,7 +173,7 @@ def test_request_with_release_number(self):
 
         release_number = self.existing_variant.Data_Release_id
         request = self.factory.get(
-            '/data/?format=json&order_by=Gene_Symbol&direction=ascending&page_size=20&page_num=0&search_term=%s&release=4&include=Variant_in_ENIGMA&include=Variant_in_ClinVar&include=Variant_in_1000_Genomes&include=Variant_in_ExAC&include=Variant_in_LOVD&include=Variant_in_BIC&include=Variant_in_ESP&include=Variant_in_exLOVD' % new_variant_1.Genomic_Coordinate_hg38)
+            '/data/?format=json&order_by=Gene_Symbol&direction=ascending&page_size=20&page_num=0&search_term=&release=%s&include=Variant_in_ENIGMA&include=Variant_in_ClinVar&include=Variant_in_1000_Genomes&include=Variant_in_ExAC&include=Variant_in_LOVD&include=Variant_in_BIC&include=Variant_in_ESP&include=Variant_in_exLOVD' % new_variant_1.Genomic_Coordinate_hg38)
```
zfisch: this is off -- it's passing the genomic coordinate in as the release number, which shouldn't work. update this test so that it asks for the release_number. you'll need to update the assertion as well so that it returns the variant with the release number expected

website/django/data/searchtest.py
```diff
@@ -224,7 +186,7 @@ def test_request_with_release_number(self):
 
         response_variant = response_data["data"][0]
 
-        self.assertEqual(response_variant["Data_Release_id"], new_variant_1.Data_Release_id)
+        self.assertEqual(response_variant["Data_Release_id"], new_variant_2.Data_Release_id)
```
zfisch: this should be something like: `self.assertEqual(response_variant["Data_Release_id"], {whatever_variant_you_ask_for}.Data_Release_id)`

website/django/data/searchtest.py
```diff
@@ -254,12 +216,12 @@ def test_request_without_release_number(self):
...
     #Filter testing: Gene type, pathogenicity, and source selection are all filter options
     def test_filter_by_gene_type_brca_1(self):
-        '''Tests filtering by 'Gene' option by BRCA1'''
+        '''Tests gene filter to ensure only variants with the gene specified in the request are returned'''
         #creates a new BRCA2 variant
-        new_variant_brca2 = test_data.new_variant()
-        new_variant_brca2["Gene_Symbol"] = "BRCA2"
-        new_variant_brca2["Genomic_Coordinate_hg38"] = "chr13:111111:A>G"
-        (new_variant_brca2, new_current_variant_brca2) = create_variant_and_materialized_view(new_variant_brca2)
+        new_variant_brca1 = test_data.new_variant()
```
zfisch: this should include the creation of 2 variants, one with gene_symbol brca1 and one with gene_symbol brca2, then it should ensure only the variants with the requested gene_symbol is returned. if there are no variants with gene_symbol brca2 in the database, the test doesn't tell us anything useful. also, shouldn't it get back two variants for brca1 since existing_variant also has a gene_symbol of brca1?

website/django/data/searchtest.py
```diff
@@ -479,22 +433,13 @@ def test_show_deleted_not_true(self):
 
         response_data = json.loads(response.content)
 
-        self.assertEqual(response_data['count'], 1)
+        #0 variants should be in response_data['count']
+        self.assertEqual(response_data['count'], 0)
```
zfisch: why isn't existing_variant returned? if the issue is that new_variant_deleted and existing_variant have the same genomic_coordinate, the genomic_coordinate for new_variant_deleted should be different or the test is confusing.

zfisch:
Never mind, looks like you already caught some of the issues I mentioned! I'll give it a once over later today and we'll see if we can get it wrapped up and merged!

PR #604, opened 2017-08-26
Cryptohash

JohnnieDM:


melissacline:
One thing: we shouldn't put large data files in GitHub, since it chews up everyone's disk space.  A better approach is to put them in Synapse, or other cloud service.  Let me know if you need some pointers in getting started with Synapse.

PR #608, opened 2017-08-31
convert ESP percentage allele frequencies to fractions in the Allele_…

zfisch:
Converts ESP allele frequencies to fractions for the aggregate Allele_Frequency field.


cmarkello:
Looks good to me!

PR #612, opened 2017-09-07
Downloads the shared lovd data, if it cannot be found locally

izcram:
Added a new task in the Shared LOVD pipeline which downloads the data from the shared LOVD server.

### DIFF
--- a/pipeline/luigi/CompileVCFFiles.py
+++ b/pipeline/luigi/CompileVCFFiles.py
@@ -726,7 +726,12 @@ class CopyEXLOVDOutputToOutputDir(luigi.Task):
 ###############################################
 
 
-class ConvertSharedLOVDToVCF(luigi.Task):
+class DownloadLOVDInputFile(luigi.Task):
+    """ Downloads the shared LOVD data
+
+    If the pipeline is run on a machine from which it is not possible to download the data (currently IP based authentication)
+    the file can be manually staged in the path of `lovd_data_file`. In this case, the task will not be run.
+    """
     date = luigi.DateParameter(default=datetime.date.today())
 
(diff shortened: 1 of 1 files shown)


pipeline/luigi/CompileVCFFiles.py
```diff
@@ -1499,7 +1524,8 @@ def requires(self):
         yield CopyG1KOutputToOutputDir(self.date, self.resources_dir, self.output_dir, self.file_parent_dir)
         yield CopyEXACOutputToOutputDir(self.date, self.resources_dir, self.output_dir, self.file_parent_dir)
         yield CopyEXLOVDOutputToOutputDir(self.date, self.resources_dir, self.output_dir, self.file_parent_dir)
-        yield CopySharedLOVDOutputToOutputDir(self.date, self.resources_dir, self.output_dir, self.file_parent_dir)
+        yield CopySharedLOVDOutputToOutputDir(date=self.date, resources_dir=self.resources_dir, output_dir=self.output_dir, file_parent_dir=self.file_parent_dir)
```
zfisch: any reason the syntax is different here? i.e. `date=self.date` instead of just `self.date`?

if it's more robust, could you make a follow up PR that makes the same change to all of the other calls as well?

izcram: Mainly for robustness reasons. I'll keep that in mind if I come around refactoring this file (e.g. splitting it up etc). But before doing that, I'd like to have a closer look at the testing infrastructure in place

zfisch: makes sense. there's currently no testing infrastructure built around compileVCFfiles.py

PR #617, opened 2017-09-12
Feature/dockerized pipeline

izcram:
Code to run pipeline in a docker container.

Needs some more fine tuning (WIP), but overall structure is there.

After the review, I'll squash the commits together, s.t. we don't have WIP commits in our master branch ;)


pipeline/docker/README.md
```diff
@@ -0,0 +1,57 @@
...
+```
+docker run -u $(id -u ${USER}):$(id -g ${USER}) \
+       -v  path_to_resource_files:/files/resources \
+       -v  path_to_output_directory:/files/data \
+       -v  optional_path_to_code_base:/opt/brca-exchange \
+       -v  path_to_pipeline_credentials.cfg:/opt/luigi_pipeline_credentials.cfg \
+       -v  path_to_previous_release_built_with_change_types.tsv:/files/previous_release/built_with_change_types.tsv \
+       -v  path_to_release_notes.txt:/files/release_notes.txt \
+       -v /tmp:/.synapseCache \
+       -e PREVIOUS_RELEASE_DATE=07-20-2017 \
```
zfisch: `PREVIOUS_RELEASE_DATE=07-20-2017` should probably be something like `PREVIOUS_RELEASE_DATE=MM-DD-YYYY`

pipeline/docker/run_luigi.sh
```diff
@@ -0,0 +1,19 @@
...
+RELEASE_NOTES=/files/release_notes.txt
+
+cd /opt/brca-exchange
+
+echo "Running brca exchange pipeline:"
+echo "Git hash: $(git log | head -n 1)"
+
+cd /opt/brca-exchange/pipeline/luigi
+
+python -m luigi --module CompileVCFFiles RunAll --resources-dir ${BRCA_RESOURCES} --file-parent-dir ${PARENT_DIR} --otuput-dir ${OUTPUT_DIR} --previous-release ${PREVIOUS_RELEASE} --previous-release-date ${PREVIOUS_RELEASE_DATE} --release-notes ${RELEASE_NOTES} --local-scheduler
```
zfisch: definitely want to include the fix for `--otuput-dir ${OUTPUT_DIR}` to `--output-dir ${OUTPUT_DIR}`

pipeline/requirements.txt
```diff
@@ -5,26 +5,17 @@ bioutils==0.1.4
...
 lockfile==0.12.2
 luigi==2.2.0
 nose==1.3.7
 numpy==1.11.1
 Parsley==1.3
-pip==8.1.2
-psycopg2==2.6.2
+pip==9.0.1
 pygr==0.8.2
-pyhgvs==0.9.4
```
zfisch: this seems odd -- i know pyhgvs is used during brca_pseudonym_generator and some other places throughout the pipeline

melissacline: We might be able to get rid of this dependency in the future, now that Reese Hart's hgvs library converts to hg38 coordinates.  The reason why we were using pyhgvs (the Counsyl library) before was that the hgvs library didn't support hg38 yet.

izcram: 0.9.4 doesn't seem to be in PyPi, therefore it is built from source instead: https://github.com/BRCAChallenge/brca-exchange/pull/617/files#diff-8f09b8970829c955182f646b3026bccbR42

Good to know that we can get rid of it, I found it quite confusing to have hgvs and pyhgvs in the same code ;)

PR #639, opened 2017-10-16
Allow various scale options for scaled allele frequency chart

strbean:
Allow the user to change the scale between various presets by clicking the scaled allele frequency chart. Toggles between 3 scales: The max of all frequencies displayed on chart, 1%, and 0.1%

I'm going to do some more QA on this, but we're hoping to get this out in the beta push next week (10/23?). Needs code review.

### DIFF
--- a/website/js/AlleleFrequencyCharts.js
+++ b/website/js/AlleleFrequencyCharts.js
@@ -5,141 +5,181 @@ var React = require('react'),
     util = require('./util'),
     _ = require('lodash');
 
-var alleleFrequencyCharts = function (variant, prop) {
-    let frequencyProps, title, pointFormat;
-    if (prop === 'Allele_Frequency_Charts_1000_Genomes') {
-        if (!variant['Variant_in_1000_Genomes']) { // eslint-disable-line dot-notation
-            return false;
+var alleleFrequencyCharts = function(variant, prop) {
+    return <AlleleFrequencyCharts variant={variant} prop={prop} />;
+};
+
+class AlleleFrequencyCharts extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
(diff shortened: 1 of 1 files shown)


website/js/AlleleFrequencyCharts.js
```diff
@@ -5,141 +5,181 @@ var React = require('react'),
...
 
-    for (let frequencyType of frequencyProps) {
-        if (!util.isEmptyField(variant[frequencyType.prop])) {
-            categories.push(frequencyType.label);
-            data.push({
-                y: parseFloat(variant[frequencyType.prop]),
-                number: variant[frequencyType.prop.replace("frequency", "number")],
-                count: variant[frequencyType.prop.replace("frequency", "count")]
-            });
+        if (_.every(_.values(_.pick(variant, _.map(frequencyProps, e => e.prop))), util.isEmptyField)) {
```
zfisch: could you add a comment explanation here to save future devs from trying to read this?

zfisch: I also think i've seen this before, if so it could be wrapped into a function with an explanation

website/js/AlleleFrequencyCharts.js
```diff
@@ -5,141 +5,181 @@ var React = require('react'),
...
-            max: chart2Max,
-            tickInterval: null,
-            labels: {
-                rotation: 45,
-                style: { fontSize: "8px" },
-                x: -2,
-                y: 2
+        }
+
+        let chart2Max = Math.max.apply(null, _.filter(_.map(_.values(_.pick(variant, _.map(frequencyProps, e => e.prop))), parseFloat), e => isFinite(e)));
```
zfisch: same, comment or function would improve readability

Issue #640, opened 2017-10-17
get previous release and date from tar.gz during pipeline run


zfisch:

zfisch:
instead of having to pass a previous release .tsv file and previous release date to the luigi run task, we should be able to simply point to a .tar.gz release and let the pipeline figure it out.

PR #641, opened 2017-10-17
changing luigi script, s.t. it can take only a previous release archi…

izcram:
…ve instead of a date and a previous data file extracted from the archive (#640)

### DIFF
--- a/pipeline/luigi/CompileVCFFiles.py
+++ b/pipeline/luigi/CompileVCFFiles.py
@@ -9,6 +9,10 @@ import luigi
 import synapseclient
 import csv
 from luigi.util import inherits, requires
+import re
+import tempfile
+import shutil
+import json
 
 from retrying import retry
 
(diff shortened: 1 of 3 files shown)


pipeline/luigi/CompileVCFFiles.py
```diff
@@ -1383,6 +1384,17 @@ def run(self):
 
 @requires(FindMissingReports)
 class RunDiffAndAppendChangeTypesToOutput(luigi.Task):
+    def _extract_release_date(self, archive_path):
```
zfisch: What do you think about pulling the release date from the output/release/metadata/version.json file rather than from the tar.gz filename? The version.json file is where data comes from that we display at http://brcaexchange.org/releases for example. In theory they should be the same, but it's possible that a user error could lead to a disparity.

izcram: Yes, good point, I didn't think of that. I'll fix it.

PR #653, opened 2017-11-13
Calculate variant priors

ccdlott:
First functions to calculate variant prior probabilities for variants in built.tsv file

### DIFF
--- a/pipeline/splicing/calcVarPriors.py
+++ b/pipeline/splicing/calcVarPriors.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python
+
+'''
+calcVarPriors
+
+Parses a tsv file (default built.tsv) containing variant information and for each variant in file 
+calculates either the prior probability of pathogenicity or a prior ENGIMA classification based on variant type and variant location
+'''
+
+import argparse
+import csv
+
+def getVarType(variant):
+    '''
+    Returns a string describing type of variant (substitution, deletion, insertion, or delins) depending on variant reference and alternate alleles
+    '''
+    if len(variant["Ref"]) == len(variant["Alt"]):
+        if len(variant["Ref"]) == 1:
+            varType = "substitution"
+        else:
+            varType = "delins"
+    else:
+        # variant is an indel
+        if len(variant["Ref"]) > len(variant["Alt"]):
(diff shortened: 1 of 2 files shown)


pipeline/splicing/calcVarPriors.py
```diff
@@ -0,0 +1,134 @@
...
+    acceptableBases = ["A", "C", "T", "G", "N", "R", "Y"]
+    badBases = 0
+    if len(sequence) > 0:
+        for base in sequence:
+            if base not in acceptableBases:
+                badBases += 1
+            if badBases == 0:
+                acceptableSequence = True
+            else:
+                # badBases > 0
```
zfisch: can rem comment

zfisch: one example of a simpler if/else block is:
```
if len(sequence) > 0:
    for base in sequence:
        if base not in acceptableBases:
            return False
    return True
else:
    return False
```
            
no need for the badBases or acceptableSequence variables

pipeline/splicing/calcVarPriors.py
```diff
@@ -0,0 +1,134 @@
...
+    if varGene == "BRCA1": 
+        varStrand = '-'
+    elif varGene == "BRCA2":
+        varStrand = '+'
+    else:
+        # varGene not BRCA1 or BRCA2
+        varStrand = ""
+
+    return varStrand
+
```
zfisch: similarly:

```
if varGene == "BRCA1": 
    return '-'
elif varGene == "BRCA2":
    return '+'
else:
    return ""
```

pipeline/splicing/calcVarPriors.py
```diff
@@ -0,0 +1,134 @@
...
+    '''
+    Returns a string describing type of variant 
+    -substitution, deletion, insertion, delins, other
+    depending on variant reference and alternate alleles
+    '''
+    varRef = variant["Ref"]
+    varAlt = variant["Alt"]
+    acceptableRefSeq = checkSequence(varRef)
+    acceptableAltSeq = checkSequence(varAlt)
+
```
zfisch: could also use returns instead of `varType` variable if desired, nbd

PR #674, opened 2017-12-05
Updated hgvs transcript reference IDs for BIC VCF processing

cmarkello:


cmarkello:
@zfisch Ready for review. It's just a simple two line change.

PR #675, opened 2017-12-07
Ensembl VEP API Implementation

ccdlott:
Please review [PR #653](https://github.com/BRCAChallenge/brca-exchange/pull/653/files) first. Only commits from Dec 5th are different from previous PR

### DIFF
--- a/pipeline/splicing/calcVarPriors.py
+++ b/pipeline/splicing/calcVarPriors.py
@@ -123,24 +123,26 @@ def getVarType(variant):
     -substitution, deletion, insertion, delins, other
     depending on variant reference and alternate alleles
     '''
-    acceptableRefSeq = checkSequence(variant["Ref"])
-    acceptableAltSeq = checkSequence(variant["Alt"])
+    varRef = variant["Ref"]
+    varAlt = variant["Alt"]
+    acceptableRefSeq = checkSequence(varRef)
+    acceptableAltSeq = checkSequence(varAlt)
     
     if acceptableRefSeq == True and acceptableAltSeq == True: 
-        if len(variant["Ref"]) == len(variant["Alt"]):
-            if len(variant["Ref"]) == 1:
+        if len(varRef) == len(varAlt):
+            if len(varRef) == 1:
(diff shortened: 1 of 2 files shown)


pipeline/splicing/calcVarPriors.py
```diff
@@ -29,11 +38,69 @@ def getVarStrand(variant):
     if varGene == "BRCA1": 
         return '-'
     elif varGene == "BRCA2":
-        return '+'
+         return '+'
```
zfisch: it's hard to see if the spacing's right here, but make sure it's the correct number!

pipeline/splicing/calcVarPriors.py
```diff
@@ -29,11 +38,69 @@ def getVarStrand(variant):
...
+
+def getVarConsequences(variant):
+    '''
+    Given a variant, uses Ensembl VEP API to get variant consequences
+    (e.g. intron variant, frameshift variant, missense variant)
+    using variant chromosome, Hg38 start, Hg38 end, and alternate allele as input for API
+    returns a string detailing consequences of variant
+    '''
+
+    server = "http://rest.ensembl.org"
```
zfisch: it's probably best to declare these variables outside the function if they're constant so they don't need to be recreated with every call. in this case, our standard practice is to write them as all caps e.g. `SERVER = "http://rest.ensembl.org"`

pipeline/splicing/calcVarPriors.py
```diff
@@ -103,13 +170,14 @@ def main():
...
-    args = parser.parse_args()
+    parser.add_argument('-b', "--boundaries", default="ENIGMA",
+                        help="Specifies which boundaries (ENIGMA or PRIORS) to use for clinically important domains")
+    args = parser.parse_args()    
     
     inputData = csv.DictReader(open(args.inputFile, "r"), delimiter="\t")
     for variant in inputData:
         varDict = getVarDict(variant)
-
+
```
zfisch: can remove whitespace

PR #684, opened 2017-12-19
adding possibility to set date explicitly in docker pipeline

izcram:
also adding vi to docker image for convenience


izcram:
@zfisch yes, it will still work properly as in this case it will take the current date, see https://github.com/BRCAChallenge/brca-exchange/blob/master/pipeline/luigi/CompileVCFFiles.py#L1514 . This is the behavior we've had so far.

PR #686, opened 2017-12-20
Determine variant location based on genomic position

ccdlott:


pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+    varExons = getExonBoundaries(variant)
+    varStrand = getVarStrand(variant)
+    varOutBounds = varOutsideBoundaries(variant)
+    if varOutBounds == False:
+        for exon in varExons.keys():
+            exonStart = varExons[exon]["exonStart"]
+            exonEnd = varExons[exon]["exonEnd"]
+            if varStrand == "+":
+                if varGenPos > exonStart and varGenPos <= exonEnd:
+                    return True
```
zfisch: careful here... any time you hit a return statement inside of a for loop it will exit the for loop (otherwise known as shortcutting the loop). Is this intentional? I noticed there are many of these in this PR that will need to be looked over carefully.

also, it seems you'd only want to shortcut the for loop with a `return False` -- i.e. if any key falls outside the exon return false, otherwise all keys in the for loop pass and you can return true.

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+    '''
+    varGenPos = int(variant["Pos"])
+    varStrand = getVarStrand(variant)
+    donorBounds = getRefSpliceDonorBoundaries(variant)
+    for exon in donorBounds.keys():
+        donorStart = donorBounds[exon]["donorStart"]
+        donorEnd = donorBounds[exon]["donorEnd"]
+        if varStrand == "+":
+            if varGenPos >= donorStart and varGenPos <= donorEnd:
+                return True
```
zfisch: same concern with returns in a for loop... please double check

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+    '''
+    varGenPos = int(variant["Pos"])
+    varStrand = getVarStrand(variant)
+    acceptorBounds = getRefSpliceAcceptorBoundaries(variant)
+    for exon in acceptorBounds.keys():
+        acceptorStart = acceptorBounds[exon]["acceptorStart"]
+        acceptorEnd = acceptorBounds[exon]["acceptorEnd"]
+        if varStrand == "+":
+            if varGenPos >= acceptorStart and varGenPos <= acceptorEnd:
+                return True
```
zfisch: return in for loop, double check

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+    '''
+    varGenPos = int(variant["Pos"])
+    varGene = variant["Gene_Symbol"]
+    inExon = varInExon(variant)
+    if varGene == "BRCA1":
+        for domain in brca1CIDomains[boundaries].keys():
+            domainStart = brca1CIDomains[boundaries][domain]["domStart"]
+            domainEnd = brca1CIDomains[boundaries][domain]["domEnd"]
+            if varGenPos <= domainStart and varGenPos >= domainEnd and inExon == True:
+                return True
```
zfisch: return in for loop, double check

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+    elif varGene == "BRCA2":
+        for domain in brca2CIDomains[boundaries].keys():
+            domainStart = brca2CIDomains[boundaries][domain]["domStart"]
+            domainEnd = brca2CIDomains[boundaries][domain]["domEnd"]
+            if varGenPos >= domainStart and varGenPos <= domainEnd and inExon == True:
+                return True
```
zfisch: check

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+    return False
+                
+
+def varInGreyZone(variant):
+    '''
+    Given a variant, determines if variant is before, in, or after the grey zone
+    Returns false if variant is before the grey zone
+    Returns "grey_zone_variant" if variant is in the grey zone
+    Return "after_grey_zone_variant" if variant is after the grey zone
```
zfisch: this function as well as the naming is confusing. It seems this function checks for a few things and could either be broken up better or renamed to show it's true purpose.

Currently the function checks if the variant gene is BRCA1 or BRCA2, if the variant is in the grey zone, and if the variant is outside the grey zone AND not in UTR, otherwise it returns False. it never returns True (which is confusing for a function with a name that implies a boolean result).

see if you can refactor this for clarity.

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+        greyZoneEnd = greyZones[varGene]["greyZoneEnd"]
+        if varGenPos >= greyZoneStart and varGenPos <= greyZoneEnd:
+            return "grey_zone_variant"
+        elif varGenPos > greyZoneEnd and inUTR == False:
+            return "after_grey_zone_variant"
+    return False
+                
+
+def getVarLocation(variant, boundaries):
+    '''
```
zfisch: again, the return statements herein will shortcut the function at any point. maybe this is intentional, but just want to double check

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+        inCiDomain = varInCiDomain(variant, boundaries)
+        if inCiDomain == True:
+            inSpliceDonor = varInSpliceDonor(variant)
+            if inSpliceDonor == True:
+                return "CI_splice_donor_variant"
+            inSpliceAcceptor = varInSpliceAcceptor(variant)
+            if inSpliceAcceptor == True:
+                return "CI_splice_acceptor_variant"
+            return "CI_domain_variant"
+        inSpliceDonor = varInSpliceDonor(variant)
```
zfisch: inSpliceDonor gets called whether inCiDomain is true or not, in this case you can move it above the inCiDomain check and not have to devote multiple lines to it.

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+            return "splice_acceptor_variant"
+        inGreyZone = varInGreyZone(variant)
+        if inGreyZone != False:
+            return inGreyZone
+        inUTR = varInUTR(variant)
+        if inUTR == True:
+            return "UTR_variant"
+        return "exon_variant"
+    else:    
+        inSpliceDonor = varInSpliceDonor(variant)
```
zfisch: again, maybe just call it at the very beginning along with inSpliceAcceptor to reduce the number of lines in this method?

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+        exonEnd = int(exonEnds[exonCount - 1])
+        if varTranscript == "NM_007294.3":
+            if exonCount >= 4:
+                # because transcript NM_007294.3 does not include an exon 4, goes from exon 3 to exon 5
+                exonName = "exon" + str(exonCount + 1)
+            else:
+                exonName = "exon" + str(exonCount)
+        else:
+            exonName = "exon" + str(exonCount)
+        varExons[exonName] = {"exonStart": exonStart,
```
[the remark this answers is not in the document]
ccdlott: It does have to be run on each iteration because it is adding the exon start and exon stop for each exon every time it goes through the loop. Unless I'm missing an easier way to do this...

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+        exonStarts = transcriptData["exonStarts"].split(",")
+        exonEnds = transcriptData["exonEnds"].split(",")
+    else:
+        # transcript is on '-' strand
+        exonStarts = list(reversed(transcriptData["exonEnds"].split(",")))
+        exonEnds = list(reversed(transcriptData["exonStarts"].split(",")))
+
+    varExons = {}
+    varExonCount = transcriptData["exonCount"]
+    exonCount = 1
```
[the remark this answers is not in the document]
ccdlott: I added a comment in the code to clarify this, but it starts at 1 because all transcripts start with exon 1, not exon 0 and I'm using the variable exonCount to directly set the exon name as it iterates through the loop. I could theoretically start it at zero, but then there would just be more +1s in the code

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+    if variant["Gene_Symbol"] == "BRCA1":
+        del varExons["exon24"]
+    elif variant["Gene_Symbol"] == "BRCA2":
+        del varExons["exon27"]
+    varStrand = getVarStrand(variant)
+    donorBoundaries = {}
+    for exon in varExons.keys():
+        exonEnd = int(varExons[exon]["exonEnd"])
+        if varStrand == "+":
+            donorStart = exonEnd - 3 + 1
```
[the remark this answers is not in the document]
ccdlott: Lol, I added some comments to hopefully explain this and part of the reason I wrote so many unittests (specifically testing boundary points) for this code was to make sure that there were no off by one errors. If it's still unclear lmk and I can elaborate further in the comments

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+            donorEnd = exonEnd + 6
+        else:
+            donorStart = exonEnd + 3
+            donorEnd = exonEnd - 6 + 1
+        donorBoundaries[exon] = {"donorStart": donorStart,
```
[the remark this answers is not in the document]
ccdlott: As the loop iterates through the exons it adds donor start and donor end position for each exon so this call needs to be made with every iteration. Again, unless I'm missing an easier way to do this...

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+    '''
+    varExons = getExonBoundaries(variant)
+    if variant["Gene_Symbol"] == "BRCA1" or variant["Gene_Symbol"] == "BRCA2":
+        del varExons["exon1"]
+    varStrand = getVarStrand(variant)
+    acceptorBoundaries = {}
+    for exon in varExons.keys():
+        exonStart = int(varExons[exon]["exonStart"])
+        if varStrand == "+":
+            acceptorStart = exonStart - 20 + 1
```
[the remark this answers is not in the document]
ccdlott: Added comments to explain, again lmk if it still is unclear and I can elaborate further

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+            acceptorEnd = exonStart + 3
+        else:            
+            acceptorStart = exonStart + 20
+            acceptorEnd = exonStart - 3 + 1
+        acceptorBoundaries[exon] = {"acceptorStart": acceptorStart,
```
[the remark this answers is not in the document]
ccdlott: as the loop iterates it adds acceptor start and acceptor end for each exon to dictionary so I want this call to be made with every iteration

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+            exonEnd = varExons[exon]["exonEnd"]
+            if varStrand == "+":
+                if varGenPos > exonStart and varGenPos <= exonEnd:
+                    return True
+            else:
+                if varGenPos <= exonStart and varGenPos >= exonEnd:
+                    return True
+        return False
+    else:
+        return "varOutBounds"
```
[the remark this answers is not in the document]
ccdlott: No it's not used for anything specific, I changed it to just return false

pipeline/splicing/calcVarPriors.py
ccdlott: I did intentionally mean to shortcut the loop. I thought it would be better for it to exit the loop as soon as it found that the genomic position was in an exon and then stop checking the rest of the exons. If it finished going through the loop then I wanted it to return false because then it meant that the genomic position was not in an exon. But I'm happy to change it if it makes more sense to do the opposite

pipeline/splicing/calcVarPriors.py
```diff
@@ -139,22 +168,314 @@ def getVarType(variant):
...
+    splice donor region = last 3 bases in exon and first 6 bases in intron
+    Returns True if variant in splice donor region
+    '''
+    varGenPos = int(variant["Pos"])
+    varStrand = getVarStrand(variant)
+    donorBounds = getRefSpliceDonorBoundaries(variant)
+    for exon in donorBounds.keys():
+        donorStart = donorBounds[exon]["donorStart"]
+        donorEnd = donorBounds[exon]["donorEnd"]
+        if varStrand == "+":
```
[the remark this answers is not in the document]
ccdlott: I could add a function, similar to the one below:
```
def checkWithinBoundaries(varStrand, varGenPos, boundaryStart, boundaryEnd):
    if varStrand == "+":
        if varGenPos >= boundaryStart and varGenPos <= boundaryEnd:
            return True
```
etc.
But I think I could really only use this on 2 functions (varInSpliceDonor and varInSpliceAcceptor). I could potentially use it in varInExon but would then have to add another argument to the function about whether to use  >=  or  > 
I'm happy to add that in if you think it would be helpful

pipeline/splicing/calcVarPriors.py
ccdlott: Again as before this was intentional, if it found that the position was in a splice donor region I wanted it to immediately exit the loop and return true so it wouldn't have to continue iterating through the rest of the splice donor regions. If it did not exit the loop then the variant would not be in a splice donor region and so the function would then return false

pipeline/splicing/calcVarPriors.py
ccdlott: Same as above comment but with splice acceptor regions instead of splice donor regions

pipeline/splicing/calcVarPriors.py
ccdlott: Same principal as above, wanted the code to exit immediately if variant position was in a clinically important domain. Again, can change if it makes more sense to do it a different way

pipeline/splicing/calcVarPriors.py
ccdlott: Same as above but for minus strand instead of plus strand transcripts

pipeline/splicing/calcVarPriors.py
ccdlott: I could also potentially use a "checkWithinBoundaries" function here but would have to then have to just add another line to check that the inExon == True before running the checkWithinBoundaries funciton

pipeline/splicing/calcVarPriors.py
ccdlott: same as above with the checkWithinBoundaries function

pipeline/splicing/calcVarPriors.py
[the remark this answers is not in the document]
zfisch: ah yea, nm

pipeline/splicing/calcVarPriors.py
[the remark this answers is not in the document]
zfisch: yep, my mistake

pipeline/splicing/calcVarPriors.py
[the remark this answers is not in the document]
zfisch: up to you! either way's fine

pipeline/splicing/calcVarPriors.py
zfisch: always best to keep it DRY (don't repeat yourself) when possible

pipeline/splicing/calcVarPriors.py
ccdlott: I split it into 2 different functions: varInGreyZone and varAfterGreyZone so hopefully it makes more sense now

pipeline/splicing/calcVarPriors.py
ccdlott: it's intentional

pipeline/splicing/calcVarPriors.py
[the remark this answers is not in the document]
ccdlott: I added the function, definitely makes it easier to read and makes so I don't repeat myself

PR #688, opened 2017-12-27
Mocks fetch_gene_coordinates responses 

ccdlott:
And cleans up unnecessary self prefixes


ccdlott:
Hopefully this will make it unnecessary to add mysqlclient and numpy to the test-requirements.txt file. Will try deleting these from PR #687 after this is merged

PR #690, opened 2017-12-29
hide empty chart rows as well as empty allele frequency groups when n…

zfisch:
…ecessary

### DIFF
--- a/website/js/index.js
+++ b/website/js/index.js
@@ -787,6 +787,7 @@ var VariantDetail = React.createClass({
                     title = "Abbreviated AA Change";
                 }
 
+                // frequency charts are not displayed if they're empty
                 if (rowDescriptor.replace) {
                     rowItem = rowDescriptor.replace(variant, prop);
                     if (rowItem === false) {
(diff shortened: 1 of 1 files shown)


zfisch:
@falquaddoomi I requested your review on this since you've also worked around this code before

website/js/index.js
```diff
@@ -787,6 +787,7 @@ var VariantDetail = React.createClass({
                     title = "Abbreviated AA Change";
                 }
 
+                // frequency charts are not displayed if they're empty
                 if (rowDescriptor.replace) {
                     rowItem = rowDescriptor.replace(variant, prop);
                     if (rowItem === false) {
```
falquaddoomi: Minor nitpick, but should the comment from line 790 be for this line instead?

falquaddoomi: Minor nitpick, but should the comment from line 790 be for this line instead?

zfisch:
@falquaddoomi great suggestions, just updated the PR accordingly. please re-review when you have a chance :)

PR #695, opened 2018-01-08
Splicing vis

strbean:
Transcript Visualization panel in variant detail. Still in progress / QA


zfisch:
@strbean Could you provide a detailed explanation of what this is and what's left to do so whoever takes it over has some context?

Specifically, they'll need to know:

1. The purpose/goals of the branch.
2. What's complete.
3. What's left to do.
4. Any unsolved problems in the code.

Thanks!

falquaddoomi:
I've added two new package.json dependencies, d3-scale (to compute the scaling domains for the segments), and immutability-helper to efficiently update some of the nested state collections that I'm using. They're both pretty small libraries.

Also, the code introduces a bunch of new components, specifically:

- *Splicing:* a set of Transcript and Zoom components, plus controls and state (e.g. which regions are visible) for the visualization.
- *Transcript/Zoom*: both are a collection of segments (exons/introns). The Zoom component only shows segments which contain the variant. The "zoom lines" that connect Transcript and Zoom belong to Transcript.
- *Intron/Exon:* both are segments with associated start and end coordinates.
- *SegmentRegions:* a component within an Intron/Exon that maps logical variants and clinically important regions to Segment-local regions instances constrained within the current segment. A single logical region may map to multiple Region instances when it crosses a segment boundary. Clinically important regions are rendered directly as Regions, whereas variant Regions are contained within a Variant instance.
- *Variant:* a wrapper for variant data that splits a variant into three discrete Regions: inserted, changed, and deleted.
- *Region:* a visible span over a Segment. Used for rendering the variant's deleted/inserted/changed spans as well as clinically important regions.

Known issues:

- The colors are still probably problematic for colorblind users. All but the "deleted variant" pattern (since it requires an SVG Pattern) are editable in the CSS; the "deleted variant" pattern is editable in the Splicing component in Splicing.js.
- Transcript and Zoom could be combined into a single component with some factoring out of common functionality, but I didn't have time to address it.

website/js/Splicing.js
```diff
@@ -0,0 +1,807 @@
...
+}
+
+class Transcript extends React.Component {
+    render() {
+        const {variant, donors, acceptors,  CIDomains, segments, width, isFlipped} = this.props;
+
+
+        // ------------------------------------------------
+        // --- precalculate scale
+        // --- TODO: scales are slightly different between exonic / intronic variants. fix.
```
zfisch: followup

website/js/Splicing.js
```diff
@@ -0,0 +1,807 @@
...
+            </g>
+        );
+    }
+}
+
+class Splicing extends React.Component {
+    constructor(props) {
+        super(props);
+
+        // const meta = geneMeta[props.variant['Gene_Symbol']];
```
zfisch: rem comment

website/js/Splicing.js
```diff
@@ -0,0 +1,807 @@
...
+
+        this.state = {
+            drawAcceptors: true,
+            drawDonors: true,
+            drawCIDomains: new Set(), // new Set(Object.keys(meta.CIDomains)),
```
zfisch: cleanup

website/js/Splicing.js
```diff
@@ -0,0 +1,807 @@
...
+                span: {
+                    start: Math.min(exon.span.start, exon.span.end),
+                    end: Math.max(exon.span.start, exon.span.end),
+                }
+            })),
+            (a) => a.span.start
+        );
+
+        // sanity check: verify that the variant actually overlaps the gene at all
+        // FIXME: should this reject variants that don't lie entirely within this gene? currently we look for just an overlap
```
zfisch: followup

website/js/Splicing.js
```diff
@@ -0,0 +1,807 @@
...
+        return (
+            <g>
+                <rect x={bpMinPx} width={widthPx} height={height}
+                    className={className}
+                    clipPath={mask && `url(#${mask})`}
+                />
+
+                {/*
+                draw the outline separately, ignoring the mask, so we can see it around variants
+                FIXME: this looks kind of weird, revisit it later perhaps
```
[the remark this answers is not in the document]
falquaddoomi: I think it's alright as it is.

For future reference, my issue is that the stroke is in the middle of the outline of the shape, and when you display something on top of the shape you can just see the part of the stroke that's outside of it (which imo looks a little odd). I'd prefer if there was a way to outset the stroke, but it seems the SVG working group is opposed to the idea -- it's been proposed in various forms, even briefly implemented as `stroke-alignment`, and then later deprecated.

website/js/Splicing.js
falquaddoomi: I believe this ended up getting fixed when I switched to d3-scale, so I'll remove the comment.

PR #698, opened 2018-01-11
Closes #697, uses an internal name field on data_releases for display…

### LINKED ISSUE #697: use a display id instead of auto incremented id for data release
currently the portal displays the release id directly from datarelease.id. we should use some other value to display to avoid any discrepencies between the internal sequence id and the user facing value.

see the previous releases page and urls with release ids for reference.

zfisch:
… purposes

Problem: Release names cannot be edited. Specifically, releases are currently named in the web portal by their id (see http://brcaexchange.org/releases), which is actually a value pulled directly from the data_release_seq_id table in postgres. Since the id is a primary key, it's extremely difficult to change. This is particularly problematic if a data deployment hits a snag -- the data_release_seq_id increments but no new data_release object is created in the database, so when a new one is created there's a missing number (e.g. the portal right now is missing Version 14).

Solution: Use a name field on data_release objects specifically for the release name. It is determined by the date the release data was generated in ascending order.

Potential outstanding issue: We use a release_id parameter in the query string to get and display specific releases in the web portal. The release_id still refers to the primary key, not the actual release name. This has the potential to confuse both users and developers.

@izcram This is most relevant to you with respect to data deployment.
@falquaddoomi This is most relevant to you with respect to user experience.

### DIFF
--- a/website/django/data/models.py
+++ b/website/django/data/models.py
@@ -9,7 +9,7 @@ class DataRelease(models.Model):
     sources = models.TextField()
     md5sum = models.TextField()
     created = models.DateTimeField(auto_now_add=True, null=True)
-    name = models.PositiveIntegerField(null=True)
+    name = models.PositiveIntegerField()
 
     class Meta:
         db_table = "data_release"
--- a/website/django/data/models.py
+++ b/website/django/data/models.py
@@ -9,9 +9,11 @@ class DataRelease(models.Model):
     sources = models.TextField()
     md5sum = models.TextField()
     created = models.DateTimeField(auto_now_add=True, null=True)
+    name = models.CharField(max_length=255)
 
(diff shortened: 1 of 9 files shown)


website/django/data/models.py
```diff
@@ -9,9 +9,11 @@ class DataRelease(models.Model):
     sources = models.TextField()
     md5sum = models.TextField()
     created = models.DateTimeField(auto_now_add=True, null=True)
+    name = models.CharField(max_length=255)
```
izcram: I'm wondering if we should set this to some integer type and not a char type. Some parts of this PR depend on the fact that a release name is an integer. I get that the release name might change to something else at some point, however, it seems safer to force another DB migration in case this happens.

PR #699, opened 2018-01-16
Calculates prior prob and ENIGMA class for variants in splice donor/acceptor regions

ccdlott:


zfisch:
Ah, yes it is a problem. Are you comfortable making your way back to where you need to be and updating the PR? If not, let me know and I can help you with it tomorrow afternoon.

ccdlott:
Done, I made the changes you requested and reverted back to a previous commit so I now only see my 8 commits for this branch

pipeline/splicing/calcVarPriors.py
```diff
@@ -369,11 +381,29 @@ def varInSpliceDonor(variant):
             return True
     return False
 
+def getVarSpliceDonorBounds(variant):
+    '''
+    Given a variant, checks if variant is in a splice donor region
+    If variant is in a splice donor, returns a dictionary with donor region boundaries where variant is located
+    '''
+    inSpliceDonor = varInSpliceDonor(variant)
+    if inSpliceDonor == True:
```
zfisch: a simpler way to write this is:

`if varInSpliceDonor(variant):`

(no need for inSpliceDonor variable)

pipeline/splicing/calcVarPriors.py
```diff
@@ -369,11 +381,29 @@ def varInSpliceDonor(variant):
...
+        varGenPos = int(variant["Pos"])
```
zfisch: this whole block from line 391-398 is also in `varInSpliceDonor`, `varInSpliceAcceptor`, and `getVarSpliceAcceptorBounds`  -- can you put it into a function to keep DRY? You'll need to take a couple arguments to make it flexible enough to handle donors and acceptors :)

ccdlott: I combined varInSpliceDonor and varInSpliceAcceptor into one function and getVarSpliceDonorBounds and getVarSpliceAcceptorBounds into another function. Is that enough to keep it DRY?

PR #701, opened 2018-01-17
don't preserve owner and group when rsyncing in deploy-dev

izcram:
Everything should have owner and user brca:brca, otherwise we keep having issues with rights.

### DIFF
--- a/deployment/deploy-dev
+++ b/deployment/deploy-dev
@@ -12,9 +12,9 @@ WEBSITE=${DEPLOYMENT}/../website
 cd ${WEBSITE}
 npm run build:prod
 
-# deploy
-rsync -a --delete --rsync-path='sudo rsync' build/ ${USER}@${HOST}:/var/www/html/beta
-rsync -a --delete --exclude="/uploads" --exclude="/downloads/*" --rsync-path='sudo rsync' django/ ${USER}@${HOST}:/var/www/backend/beta/django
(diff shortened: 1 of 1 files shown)


deployment/deploy-dev
```diff
@@ -12,9 +12,9 @@ WEBSITE=${DEPLOYMENT}/../website
 cd ${WEBSITE}
 npm run build:prod
 
-# deploy
-rsync -a --delete --rsync-path='sudo rsync' build/ ${USER}@${HOST}:/var/www/html/beta
-rsync -a --delete --exclude="/uploads" --exclude="/downloads/*" --rsync-path='sudo rsync' django/ ${USER}@${HOST}:/var/www/backend/beta/django
+# deploy (not preserving owner/group)
+rsync -rlptD --delete --rsync-path='rsync' build/ ${USER}@${HOST}:/var/www/html/beta
```
zfisch: curious -- how does this work?

I noticed in the docs ` -p, --perms                 preserve permissions`

izcram: `-p` are the file permissions like `rwxr-xr-x`.
According to the man page `-a` corresponds to `-rlptgoD`. But since keeping the owner and group creates some havoc, I've removed the `go` part.

zfisch: 👍

Issue #512, opened 2017-05-23
per report UI view


zfisch:

falquaddoomi:
Commit f31f8daa4779455c69f8f2b3b7e8d095c0187f44 introduces (hopefully?) properly-formatted fields to the reports section. Fields that have no registered display label are displayed as "?:<field>", which causes the column to overflow for particularly long ones, but there shouldn't be any of these left when the feature is completed.

Other things that remain to be done:

- Determining where to find the submitter value reliably across different sources
- Adding star ratings for each submission
- Displaying summary data in the report's header when it's collapsed
- Determining which columns from the reports should be displayed in the output
- Possibly adding controls for toggling all report visibility, or toggling on a per-source basis

zfisch:
@melissacline @falquaddoomi 

We should figure out how to handle links to the help page -- as faisal pointed out to me, each field in the per submission tiles links to the top of the help page but the field may not be included anywhere on the help page. Additionally the per submission tiles are the only tiles without question mark links in the upper right corner of the tile.

A couple options:

1. Remove links from all per submission fields until we have a chance to determine which ones we want.
2. Remove links from all per submission fields that aren't on the help pages, hook up the ones that are.
3. Determine which fields should have help content, update help content and links for those fields.

Thoughts?

melissacline:
My leaning is toward Option 3.  Amy has the first cut of the help pages in good shape for the expert-reviewed portal, and it's a good time for her to start looking at the all public data portal (which the GCs will be using frequently, so her focus groups would apply well here).  When I talk to her tomorrow, we'll talk about starting to do this.

melissacline:
Just to follow up, it will probably be a few weeks before Amy can get to this.  So if it's not too late to do #2 for now, let's do that now and plan that Amy will do a more comprehensive update in a month or so.

zfisch:
@falquaddoomi Will you have time to implement this?

> Remove links from all per submission fields that aren't on the help pages, hook up the ones that are.

falquaddoomi:
Sure thing. So far I've added an extra prop to the per-report meta, "helpKey", that explicitly specifies which ID to scroll to when the field label is clicked. If "helpKey" unspecified, then the field label renders without an underline and help icon on hover.

[...]

PR #711, opened 2018-01-24
Individual reports in nested tiles

falquaddoomi:
Implements display of individual reports on a per-source basis in nested, collapsible tiles. Currently only ClinVar and LOVD sources show individual reports. Factored out field normalization and other associated methods into a 'util' module, since it's used in both the variant details rendering and report rendering.

Also includes the addition of a '/variant/<id>/reports' endpoint in the backend, which returns the reports for a given variant.


falquaddoomi:
Just FYI, I pulled master into this branch shortly before generating the PR and resolved the conflicts manually, so it should merge cleanly.

website/js/VariantTable.js
```diff
@@ -95,24 +97,62 @@ const researchModeGroups = [
...
+            cols: [
+                // displayed here in full since the display in the header is potentially truncated
+                {title: 'Submitter(s)', prop: 'Submitters_LOVD'},
+
+                // added this field (redundantly?) b/c it's not noticeable in the header
+                {title: 'Variant Effect', prop: 'Variant_effect_LOVD'},
+
+                {title: 'Genetic Origin', prop: 'Genetic_origin_LOVD'},
+                {title: 'Individuals', prop: 'Individuals_LOVD'},
+                {title: 'Submission ID', prop: 'DBID_LOVD'}, // possibly BX_ID_LOVD, DBID_LOVD?
```
zfisch: DBID is correct -- BX_ID's are brca_exchange ids for internal purposes only. can remove comment

website/js/components/SourceReportsTile.js
```diff
@@ -0,0 +1,172 @@
...
+                    style={{cursor: 'pointer'}}>
+                    <i className="fa fa-angle-double-down" aria-hidden="true" />
+                </a>
+            </h3>
+        );
+
+        // checks if we either have no submissions for this source, or if every field in every submission is empty
+        // (consideration: it's very unlikely that we're ever actually all empty, so should we skip this?)
+        // (use the cache if it's available, or else recalculate)
+        const allEmpty = this.allEmptyCached != null ? this.allEmptyCached : this.calculateAllEmpty(submissions);
```
zfisch: theres a lot to review here so it's hard to wrap my head around this completely -- is this just a check so that if there are no submissions or if all fields of a submission are empty we can refrain from rendering empty content on the page?

it's safe to say that for any variant with ClinVar or LOVD as a source, they will have submissions (reports) from the respective source. but, for example, if a variant has ClinVar as a source, it may not have submissions from LOVD and vice versa. it should also be safe to assume that if a variant has a submission, that submission will not be all empty fields.

please let me know if this doesn't address your concern!

website/js/components/VariantSubmitter.js
```diff
@@ -0,0 +1,186 @@
...
+import classNames from 'classnames';
+import util from '../util';
+import KeyInline from './KeyInline';
+
+function sentenceCase(str) {
+    return str.replace(/\b\S/g, (t) => t.toUpperCase() );
+}
+
+// from https://www.ncbi.nlm.nih.gov/clinvar/docs/review_status/#revstat_def
+const marksToReviewStatuses = {
```
zfisch: did we ever get a consensus on whether or not stars will be included in the UI?

website/js/components/VariantSubmitter.js
```diff
@@ -0,0 +1,186 @@
...
+                }
+                </div>
+            </div>
+        );
+    },
+
+    render: function() {
+        let styles = this.getCollapsableClassSet();
+
+        // ----
```
zfisch: ??

website/js/index.js
```diff
@@ -168,7 +172,7 @@ var Home = React.createClass({
                     {logoItems}
                 </Row>
                 {this.state.showModal && <Modal bsSize="large" onRequestHide={() => this.setState({ showModal: false })}>
-                    <iframe className="vimeo-video" src="https://player.vimeo.com/video/199396428" className="vimeo-video" frameBorder="0" webkitallowfullscreen mozallowfullscreen allowFullScreen></iframe>
+                    <iframe className="vimeo-video" src="https://player.vimeo.com/video/199396428" frameBorder="0" webkitallowfullscreen mozallowfullscreen allowFullScreen></iframe>
```
zfisch: nice

website/js/VariantTable.js
```diff
@@ -95,24 +97,62 @@ const researchModeGroups = [
...
+            submitter: {title: 'Submitter', prop: 'Submitter_ClinVar'},
+            cols: [
+                // displayed here in full since the display in the header is potentially truncated
+                {title: 'Submitter', prop: 'Submitter_ClinVar'},
+
+                // added these fields (redundantly?) b/c they're not noticeable in the header
+                {title: 'Clinical Significance', prop: 'Clinical_Significance_ClinVar'},
+                {title: 'Date Last Updated', prop: 'Date_Last_Updated_ClinVar'},
+
+                {title: 'Submission Type', prop: 'Method_ClinVar'}, // likely Method_ClinVar
```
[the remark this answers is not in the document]
falquaddoomi: Right, sorry, forgot to remove some of the comment cruft. I'll remove it ASAP.

website/js/VariantTable.js
```diff
@@ -95,24 +97,62 @@ const researchModeGroups = [
...
+                {title: 'SCV Accession', prop: 'SCV_ClinVar'},
+                {title: 'Summary Evidence', prop: 'Summary_Evidence_ClinVar'},
+                {title: 'Supporting Observations', prop: 'Description_ClinVar'}, // possibly Description_ClinVar
```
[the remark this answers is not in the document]
falquaddoomi: Same, it's not needed.

website/js/components/SourceReportsTile.js
falquaddoomi: It does, thanks; if there's never a case in which a submission has all empty fields, then that check is unnecessary. I've removed it and will push an update in a minute or two.

website/js/components/VariantSubmitter.js
falquaddoomi: Heh, yeah, another forgotten "marker" comment...

website/js/index.js
falquaddoomi: 👍

falquaddoomi:
Fixed a bug with field labels in the reports not linking to the help at all. In response to #512, made the help target for each report field label configurable (via the 'helpKey' field) in the report binding meta. Currently no fields have a helpKey specified, since an issue with duplicate IDs on the current help page is preventing me from properly linking them.

website/js/components/VariantSubmitter.js
falquaddoomi: I don't recall. Also, do you mean whether it should be a star glyph or not, or whether we should be including a rating at all?

melissacline: We should definitely include the rating, that much is clear.  In terms of whether or not to represent it as a star glyph, that seems less controversial than I thought.  ClinVar has SCV (report level) and RCV (variant level) review statuses, which they refer to with the same terms.  I'd been concerned that the stars in ClinVar were specific to the RCV data, and the RCV-level stars are readily-apparent ones, but having SCV (report) level stars doesn't seem misleading.

zfisch: sounds like we can keep it as is for now

PR #714, opened 2018-01-26
Speeds up unittests

ccdlott:
and fixes path to brca.zscore.json


ccdlott:
Can confirm that all tests passed from root directory on my local machine

zfisch:
Approved! But let's wait until after the production deploy (scheduled for monday) in case we run into any issues with the tests running in circleCI.

PR #715, opened 2018-01-26
Cleanup help and faq

zfisch:
Adds panels for FAQ's with smoother animations. Allows multiple questions to be expanded at the same time.


melissacline:
@zfisch This looks so nice!

Issue #718, opened 2018-01-30
navigation items wrap to a second line when below 1200px width


falquaddoomi:
As of d722d147ea8638b08f28d7a984b3b3e727ad52eb, navigation items in the header wrap to a second line when the viewport drops below 1200px width, and persist (despite the navbar branding changing layout) until 768px, when the navigation items collapse into a dropdown menu.

falquaddoomi:
This is probably a result of adding the "FAQ" item to the navbar.

zfisch:
Nice find, think you can put in a PR for a fix in the next day? If not I can take care of it, but want to get it before release.

falquaddoomi:
Sure thing; I'll work on it tomorrow morning and hopefully have a fix within an hour or two.

falquaddoomi:
Fixed in PR #720.

PR #720, opened 2018-01-30
fixes navbar wrapping with introduction of new 'FAQ' item

falquaddoomi:
Fixes issue #718 by reducing the width and margins of the logo, as well as slightly tweaking its collapsed-state width. In the future, we might consider switching to a less fragile layout system (Flexbox?) that would be easier to adapt to changes in the number of navigation items.

### DIFF
--- a/website/js/NavBarNew.js
+++ b/website/js/NavBarNew.js
@@ -51,17 +51,24 @@ var NavBarNew = React.createClass({
     render: function () {
         var {path} = this.props;
         var brand = (
-            <a className="navbar-brand navbar-custom-spacing" href="/">
-                <h1>
-                    <span className="logo-img">
-                        <img src={brcaHeaderLogo} height="40" />
-                    </span>
-                    <span className="BRCA">BRCA</span>
-                    <span className="exchange"> Exchange</span>
-                </h1>
-                {this.props.mode === 'research_mode' && <span id="research-label" className="label label-info">All Public Data</span>}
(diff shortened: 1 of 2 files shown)


zfisch:
I don't think this is doing what it should (unless I'm running into some sort of caching issue)... I sent a video via slack to show the behavior of this branch versus the correct behavior on production.

falquaddoomi:
I think it's likely that it's a caching issue, since the video you sent appears fine for me. Here's a video that demonstrates the behavior I'm seeing locally.

out

zfisch:
out

falquaddoomi:
I'm noticing the same behavior on the Variant Details page of the production site (http://brcaexchange.org/variant/68606) using an iPhone 8 preset w/Chrome's mobile view. It looks to me like the version history table is stretching out the page; it's 544px whereas the viewport is 375px. Part of the problem is that the date column doesn't wrap (my error, introduced a while ago when we were fixing up the look & feel), but even if that column is set to wrap it still exceeds the width.

Perhaps this should be a separate issue specifically about the page being stretched past the viewport in mobile devices, since I'm noticing this behavior on other pages as well. For instance, http://brcaexchange.org/variants?search=c.956 stretches past the viewport due to the "matching variants" textbox being too wide.

melissacline:
Yes, let's make a separate issue for the page being stretched past the viewport, and that can be addressed separately.  Thanks, @falquaddoomi !

melissacline:
Ah, I see the issue has already been created.  Thanks, @zfisch !

Issue #722, opened 2018-02-02
Fix page extending beyond borders on small screens


zfisch:
I'm noticing the same behavior on the Variant Details page of the production site (http://brcaexchange.org/variant/68606) using an iPhone 8 preset w/Chrome's mobile view. It looks to me like the version history table is stretching out the page; it's 544px whereas the viewport is 375px. Part of the problem is that the date column doesn't wrap (my error, introduced a while ago when we were fixing up the look & feel), but even if that column is set to wrap it still exceeds the width.

Perhaps this should be a separate issue specifically about the page being stretched past the viewport in mobile devices, since I'm noticing this behavior on other pages as well. For instance, http://brcaexchange.org/variants?search=c.956 stretches past the viewport due to the "matching variants" textbox being too wide.

I have some ideas on how to fix these issues (hiding the "Changes" column, allowing the error text to wrap, etc.), so I'd be happy to work on it on a separate issue branch.

falquaddoomi:
Fixed in #724.

PR #724, opened 2018-02-02
Fix narrow viewport overflow #722

falquaddoomi:
Addresses #722. Allows the widths of certain elements that were previously stretching the page past the viewport on narrow displays to collapse. Specifically, the following pages' elements have been modified:

- **Variant Search:**
  - the label indicating the number of results, duplicates, etc. has had the "label" class removed, which was disallowing it from wrapping. Relevant rules have been copied into its own class, "matched-variant-count".
  - the results table previously had a few pixels of space between the table's border and its content. Moving the "table-responsive" class from its parent Col element to a new wrapping div resolved the left padding, but oddly not the right padding, although it's less noticeably since the user has to scroll to see it.
- **Variant Details:**
  - the version history table now has the "table-responsive" class, which causes it to scroll internally (rather than stretch the page) if its content is wider than the page can accommodate. I'm very mildly concerned that users might not be able to find the "changes" column if they don't realize the version history table scrolls, although I think it's better than the menu button being obscured.

### DIFF
--- a/website/js/DataTable.js
+++ b/website/js/DataTable.js
@@ -359,17 +359,19 @@ var DataTable = React.createClass({
                     </Col>
                 </Row>
                 <Row>
-                    <Col id="data-table-container" className="table-responsive" sm={12}>
-                        <FastTable
-                            className={cx(className, "table table-hover table-bordered table-grayheader")}
-                            dataArray={data}
-                            columns={renderColumns}
-                            keys={this.props.keys}
-                            buildRowOptions={this.props.buildRowOptions}
-                            onRowClick={this.props.onRowClick}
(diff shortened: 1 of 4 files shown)


falquaddoomi:
Thanks! 👍 

I did a quick review and found that the Data Releases page had the same issue. The fix for that one was a little more involved -- for some reason, "table-responsive" cuts off the edge of a bordered table between 768px and 992px, despite handling it at lower resolutions (at higher it's not truncated at all). To fix it for now I wrote a little stub table class that renders all but the outer border and rendered the responsive parent with a border, but it should probably be revisited to see if there's a way to do it entirely within Bootstrap.

PR #728, opened 2018-02-07
Fix genome track update

zfisch:
solves an issue where the relative path to the genome browser update script did not point to the correct location -- ensures absolute paths are used instead of relative paths


zfisch:
Yes and yes. Updated for your review :)

PR #729, opened 2018-02-07
Calculate prior prob for variants after grey zone

ccdlott:
Also adds global variables for transcript data so API call is not necessary for every variant

### DIFF
--- a/pipeline/splicing/calcVarPriors.py
+++ b/pipeline/splicing/calcVarPriors.py
@@ -264,7 +264,6 @@ def getExonBoundaries(variant):
     '''
     varTranscript = variant["Reference_Sequence"]
     transcriptData = getTranscriptData(varTranscript)
-    
     # parse exon starts and exon ends
     transcriptData["exonStarts"] = re.sub(",(\s)*$", "", transcriptData["exonStarts"])
     transcriptData["exonEnds"] = re.sub(",(\s)*$", "", transcriptData["exonEnds"])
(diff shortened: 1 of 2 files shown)


pipeline/splicing/calcVarPriors.py
```diff
@@ -245,7 +259,10 @@ def getExonBoundaries(variant):
     Uses function implemented in calcMaxEntScanMeanStd to get data for variant's transcript
     '''
     varTranscript = variant["Reference_Sequence"]
-    transcriptData = fetch_gene_coordinates(str(varTranscript))
+    if varTranscript == BRCA1_RefSeq:
```
zfisch: the three identical blocks can all be a function call, e.g. something like:
```
def getTranscriptData(referenceSequence):
    if varTranscript == BRCA1_RefSeq: 
            return brca1TranscriptData
    elif varTranscript == BRCA2_RefSeq:
            return brca2TranscriptData
```

then just call it when you need it:
```
transcriptData = getTranscriptData(varTranscript)

**Comment by ccdlott** [comment]
@zfisch can I merge this now? Or do I need to wait until the PR is approved before merging?

File: pipeline/splicing/calcVarPriors.py [hunk 2]
```diff
@@ -245,8 +263,8 @@ def getExonBoundaries(variant):
     Uses function implemented in calcMaxEntScanMeanStd to get data for variant's transcript
     '''
     varTranscript = variant["Reference_Sequence"]
-    transcriptData = fetch_gene_coordinates(str(varTranscript))
-
+    transcriptData = getTranscriptData(varTranscript)
+
```
zfisch: can rem white space

zfisch:
@ccdlott wait for approval unless the change is crucial or blocking and without error -- best to be diligent as much as possible

PR #732, opened 2018-02-09
init file with splice data for visualizations

zfisch:
@falquaddoomi this is all the information you'll need to get started with ui development. if we end up developing a more advanced solution for integrating other genes we can follow the same format or a very similar one and remove this file.

you'll notice the brca1 exons/regions start before they stop, but for visualization purposes i think it only matters if they're on or in the clinically important range rather than where it technically starts or stops. @melissacline is this correct?


falquaddoomi:
Looks good. Minor point, but would it be possible to name all the individual keys in each region the same way, e.g. 'start'/'end' vs. 'donorStart'/'donorEnd'? That information's already implied by which collection the regions are coming from, which would save us some code where we figure out which keys are for which collection. I'm happy to make that change myself, but I wanted to make sure there wasn't some case later on where we'd want to use individual regions outside of their collection, in which case the typing info might be helpful.

zfisch:
@falquaddoomi makes sense -- let me know if this works for you or feel free to make any needed adjustments :)

zfisch:
Regarding start and end coordinates for BRCA1 -- Chloe summed this up nicely:

"BRCA1 is transcribed on the minus strand, meaning it is read from right to left instead of the usual left to right. Genomic coordinates are always based on the plus strand so the start coordinate for BRCA1 occurs to the right of the end coordinate in genomic position (so it has a greater value)"

My understanding is that we should show the visualization in the coding direction with exons always increasing in number from left to right, despite the fact that the genomic positions would be greater on the left side of the brca1 transcript in this orientation.

PR #733, opened 2018-02-12
Update readme

zfisch:
Updates readme based on @ccdlott's feedback. Also updates steps to get data into the database.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -35,7 +35,7 @@ The server runs on Django with postgres so install and set up those
    * If this doesn't work, figure out how to set user `postgres`'s password to `postgres`.
 * **Install the python dependencies**
    * `cd website`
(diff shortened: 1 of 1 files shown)


website/README.md
```diff
@@ -15,23 +15,27 @@ The build is based on npm and webpack.
...
+   * `sudo -u postgres createdb storage.pg`
+   * If this doesn't work, figure out how to create a db called `storage.pg` for user `postgres`.
 * **Set the postgres role's password**
    * `sudo -u postgres psql postgres`
    *  at the prompt type `\password postgres` to set the password to `postgres`
+   * If this doesn't work, figure out how to set user `postgres`'s password to `postgres`.
 * **Install the python dependencies**
    * `cd website`
-   * `pip install -qU -r requirements.txt`
+   * `pip install -qU -r requirements.txt`, if you encounter failures with psycopg2, try `pip install psycopg2==2.7.3.2`
```
ccdlott: Do you want to leave the -q option here? Or would it be more helpful to say something like "if having problems with instillation use `pip install -U`  instead of `pip install -qU`"

PR #748, opened 2018-03-19
Circleci2.0

izcram:
Circleci 2.0 configs.

The deployment part of circle CI 1.0 has been removed to avoid interference with the 2.0 version, but could in principle be removed.

Regarding the deployment part for the 2.0 version, I'm not 100% confident it will work as expected, as I was not really able to test the branches and tag filters mechanism. We'll need to see it in action and make adjustments as we go....


website/test-requirements.txt
```diff
@@ -0,0 +1,6 @@
+enum34==1.1.6
```
zfisch: now that the website/test-requirements are separate from the project/test-requirements, can these requirements be cleaned up?

otherwise, would it make sense to just keep them as they are in a single test-requirements file?

izcram: I've cleaned up a bit. I think they should be kept separate, as the two parts are very different in principle.

zfisch:
@izcram it might be a good idea to merge this before we merge the reportdiff code so we can make sure CI works as expected

PR #751, opened 2018-03-20
Report versioning

zfisch:
This PR includes the following changes:

- Determines diffs between clinvar reports using a previous data release as part of the automated pipeline.
- Stores report diff information in the database.
- Fixes a bug where the wrong date updated field was being pulled from clinvar.
- Includes code to display clinvar report diffs in the UI, which has been commented out until the UI design has been properly vetted.


pipeline/clinvar/clinvar.py
```diff
@@ -174,7 +174,7 @@ def __init__(self, element, debug=False):
         self.dateLastUpdated = None
         cs = element.find("ClinicalSignificance")
         if cs != None:
-            self.dateLastUpdated = cs.get("DateLastEvaluated")
+            self.dateLastUpdated = cva.get("DateUpdated")
```
zfisch: we determined that the DateUpdated field is the correct field for determining updates to individual clinvar submissions

pipeline/utilities/releaseDiff.py
```diff
@@ -559,6 +574,23 @@ def round_sigfigs(num, sig_figs):
...
+    if isReport:
+        # reports have different identifiers based on their source
+        if type(obj) is list:
+            source = obj[0]
+        elif type(obj) is dict:
+            source = obj["Source"]
+        if source == "ClinVar":
+            return "SCV_ClinVar"
+        else:
+            # ignore non clinvar reports
```
zfisch: eventually we'll include lovd submissions as well

website/django/data/migrations/0026_report_change_type.py
```diff
@@ -0,0 +1,27 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('data', '0025_auto_20180116_1508'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            """
+            INSERT INTO data_changetype VALUES (7, 'none');
```
zfisch: variants are currently only added to the database if they have a changetype value (variants without changetypes are already in the database and therefore don't need to be added again)

unlike variants, reports may be added to the database even if they're identical to existing reports, hence the need for the 'none' changetype

website/django/data/views.py
```diff
@@ -98,28 +98,18 @@ def variant(request):
 def variant_reports(request, variant_id):
     variant_id = int(variant_id)
     query = Report.objects.filter(Variant_id=variant_id)
-
     '''
-    Uncomment the following code to filter enigma/bic submissions to clinvar/lovd
-
-    # filter out enigma submissions to clinvar/lovd and bic submissions to clinvar
-    filtered_query = []
+    NOTE: Uncomment to send report versions to UI
```
zfisch: we opted to not send reportdiffs to the UI for now (needs further review), hence the commented out code here.

website/js/index.js
```diff
@@ -873,7 +892,58 @@ var VariantDetail = React.createClass({
             );
         });
 
-        const diffRows = this.generateDiffRows(cols, data);
+        // generates variant diff rows
+        const diffRows = this.generateDiffRows(cols, data, false);
+
+        /* NOTE: Uncomment to display clinvar report diffs in UI
```
zfisch: similarly, we opted to not display reportdiffs for now

zfisch:
@izcram I requested your review on this PR since it includes some migrations and changes to data deployment. It will also introduce you a little bit to how we take data from the pipeline and display it on the web portal.

website/django/data/migrations/0026_report_change_type.py
izcram: Would be nice to integrate this comment in the code.

pipeline/luigi/CompileVCFFiles.py
```diff
@@ -1447,6 +1447,60 @@ def run(self):
...
 
 @requires(RunDiffAndAppendChangeTypesToOutput)
+class RunDiffAndAppendChangeTypesToOutputReports(luigi.Task):
+    def _extract_release_date(self, version_json):
+        with open(version_json, 'r') as f:
+            j = json.load(f)
+            return datetime.datetime.strptime(j['date'], '%Y-%m-%d')
+
+
+    def _extract_file(self, archive_path, tmp_dir, file_path):
```
izcram: This function and the above could probably be easily refactored out, as it is also used in `RunDiffAndAppendChangeTypesToOutput`

pipeline/luigi/CompileVCFFiles.py
```diff
@@ -1447,6 +1447,60 @@ def run(self):
...
+        return {'reports_with_change_types': luigi.LocalTarget(release_dir + "reports_with_change_types.tsv"),
+                'removed_reports': luigi.LocalTarget(diff_dir + "removed_reports.tsv"),
+                'added_reports': luigi.LocalTarget(diff_dir + "added_reports.tsv"),
+                'added_data_reports': luigi.LocalTarget(diff_dir + "added_data_reports.tsv"),
+                'diff_reports': luigi.LocalTarget(diff_dir + "diff_reports.txt"),
+                'diff_json_reports': luigi.LocalTarget(diff_dir + "diff_reports.json"),
+                'README': luigi.LocalTarget(diff_dir + "README.txt")}
+
+    def run(self):
+        release_dir = self.output_dir + "/release/"
```
izcram: maybe for a next time: I think it's possible to work with these Target objects returned by `self.output()`. In case of a single return, one can just do `self.output().path` to refer to the path. In this case, would need to lookup the target objects in the dictionary. This way, path logistics are not duplicated  
According to http://luigi.readthedocs.io/en/stable/api/luigi.local_target.html directories could also be created, if necessary.

zfisch: sounds like a nice cleanup option for future development :) I'm going to omit this for now

PR #753, opened 2018-03-22
Circle2.0 fixes

izcram:
Issue was, that in each workflow job it started from a fresh image, so in the deployment steps would need to redo `npm install`. There might be away around this by reusing the container from the build step.
However, going away from workflows as it seems a bit an overkill and currently cannot be tested locally (see note in https://circleci.com/docs/2.0/local-cli/#overview)


izcram:
merging now, to test if everything works as expected.

PR #774, opened 2018-04-20
Allele frequency nesting

zfisch:
Nests allele frequency fields and introduces expand/collapse functionality.

ExAC graph is only field expanded by default.


zfisch:
Good eye! I added the missing border :)

PR #776, opened 2018-04-20
Mupit

zfisch:
Adds cravat / mupit structures with links to mupit.

Mupit structures are determined in 2 places:

1. During the pipeline run as part of the automated luigi pipeline.
2. Manually using `manage.py update_mupit_structures`. This is used to update the structures of variants already in the database. It can also be run at any point in case changes are made on the MuPIT side and we need to update variants already stored in our database.


website/django/data/utilities.py
```diff
@@ -24,10 +28,104 @@ def update_autocomplete_words():
...
+def isMissenseSubstitution(ref, alt):
+    bases = ['a', 'c', 't', 'g']
+    ref = ref.lower()
+    alt = alt.lower()
+    if ref in bases and alt in bases and len(ref) == 1 and len(alt) == 1:
+        return True
+    return False
+
+
+def update_mupit_structure_for_existing_variants():
```
izcram: Is this a one time thing, or would we need to repeat it at every release? For the first case it would be could to clearly mark this, e.g. by putting all the relevant code into a separate file.

pipeline/data_merging/getMupitStructure.py
```diff
@@ -0,0 +1,134 @@
...
+        alt = variant[altIndex]
+
+        # only concerned with missense substitutions
+        if isMissenseSubstitution(ref, alt):
+            # Add empty data for each new column to prepare for data insertion by index
+            for i in range(len(new_columns_to_append)):
+                variant.append('-')
+
+            retries = 5
+            if (pos >= 32356427 and pos <= 32396972) or (pos >= 43045692 and pos <= 43125184):
```
izcram: It would be good to put these numbers into some separate module, s.t. we use consistent boundaries.

pipeline/data_merging/getMupitStructure.py
```diff
@@ -0,0 +1,134 @@
...
+                mupit_structure = get_brca_struct(chrom, pos)
+                if mupit_structure == "retry":
+                    if retries > 0:
```
izcram: There is probably some while or something similar missing somewhere. Maybe I'm missing something, but retries are not happening.

pipeline/data_merging/getMupitStructure.py
```diff
@@ -0,0 +1,134 @@
...
+        print e
+        return "retry"
+    d = json.loads(r.text)
+    structures = d['structures']
+    main_struct = None;
+    min_pref_level = sys.maxsize # max size integer
+    max_no_query_pos = -1
+    max_no_res = -1
+    pref_level_hit = False
+    for structure_id in structures:
```
izcram: Perhaps some small documentation/reference would be useful to understand what is going on in the code below.

pipeline/data_merging/getMupitStructure.py
```diff
@@ -0,0 +1,134 @@
...
+    parser = argparse.ArgumentParser(description='Determine correct BRCA structure from MuPIT.')
+    parser.add_argument('-i', '--input', type=argparse.FileType('r'),
+                        help='Input variants.')
+    parser.add_argument('-o', '--output', type=argparse.FileType('w'),
+                        help='Output variants.')
+    options = parser.parse_args()
+    return options
+
+
+def isMissenseSubstitution(ref, alt):
```
izcram: This should probably be called `isPointSubstitution` (or maybe something else) as a substitution at one position doesn't imply a missense substitution

zfisch:
NOTE: don't merge until after april/may production release

pipeline/data_merging/getMupitStructure.py
zfisch: added comment at the beginning of the function to explain how it works

PR #806, opened 2018-06-13
Fix/make hgvs protein mapping more robust

izcram:
Exceptions raised during the protein mapping are now all written to a log, as opposed to the previous behaviour, where the brca_pseudonym_generator.py script just crashed in some cases.

Also using hgvs 1.1.2 now.

### DIFF
--- a/pipeline/docker/Dockerfile
+++ b/pipeline/docker/Dockerfile
@@ -51,7 +51,7 @@ RUN rm -r /opt/leiden /opt/hgvs /root/.cache
 ARG FORCE_REBUILD=0
 ARG BRCA_GIT_REPO=https://github.com/BRCAChallenge/brca-exchange.git
 ARG BRCA_EXCHANGE_COMMIT=master
-#RUN git clone ${BRCA_GIT_REPO} && cd brca-exchange && git checkout ${BRCA_EXCHANGE_COMMIT}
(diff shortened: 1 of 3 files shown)


pipeline/docker/Dockerfile
```diff
@@ -51,7 +51,7 @@ RUN rm -r /opt/leiden /opt/hgvs /root/.cache
 ARG FORCE_REBUILD=0
 ARG BRCA_GIT_REPO=https://github.com/BRCAChallenge/brca-exchange.git
 ARG BRCA_EXCHANGE_COMMIT=master
-RUN git clone ${BRCA_GIT_REPO} && cd brca-exchange && git checkout ${BRCA_EXCHANGE_COMMIT}
+#RUN git clone ${BRCA_GIT_REPO} && cd brca-exchange && git checkout ${BRCA_EXCHANGE_COMMIT}
```
zfisch: is this an intentional change?

pipeline/requirements.txt
```diff
@@ -1,14 +1,15 @@
 backports-abc==0.4
 beautifulsoup4==4.5.1
```
zfisch: fingers crossed these updates are ok -- I'm experimenting with running the pipeline start to finish to make sure it doesn't cause problems prior to the brca_pseudonym_generator

PR #813, opened 2018-06-29
Hover-over Tooltips, Better Help Linking

falquaddoomi:
This PR introduces hover-over tooltips on the Variant Details page which are scraped from the rendered help markdown. Specifically, `<li>`s containing `<h4>`s are retrieved, then the first element following the `<h4>` is used as the tooltip text (and potentially unwrapped from its surrounding `<p>` tag if present).

The markdown parser has been modified to look for special "explicit ID" tags of the form `((id-here))` on the same line as a header declaration. If found, the tag is removed from the output and its contents are assigned as the ID of the header. If not found, the slugified header text is used as the ID, as before. (Note that you must relaunch webpack for this to take effect, as markdown-it performs preprocessing at launch.)

The tooltip-detectable regions and explicit IDs look like this in Markdown:
```
* #### Clinical Significance ((Clinical_Significance_ENIGMA))
   The first paragraph is taken as the tooltip.

   Additional paragraphs (e.g. this one) are not used, but are still visible in the help.
```

The headers in the help text that correspond to fields were given IDs corresponding to their DB column names, resolving the previous issue with ambiguous header IDs in the help text. Fields on the Variant Details page are now linked via these DB column names instead of their visible text.

This PR also includes various look-and-feel modifications to the help text; for instance, the research-mode help's table-of-contents is now linked to the internal headers in the document. Margins have been added to headers and lists to improve readability. The Help component's scroll-to-fragment code has also been extracted into its own method, which is passed to RawHTML to allow the document-internal links' behaviors to match Variant Details' linking behavior.

In the process of reviewing the help text, I found references to a few LOVD report fields ("created date", "edited date") which for some reason we weren't displaying in the UI, as well as two fields that were out of order (the allele frequency "southeast asian" and "other" fields). These have been rectified.


falquaddoomi:
Changes are still incoming, so please don't approve the request yet.

falquaddoomi:
Ok, all the changes are in; it's ready for review now.

website/js/VariantTable.js
```diff
@@ -180,8 +180,8 @@ const researchModeGroups = [
                     {title: 'EAS Allele frequency (ExAC minus TCGA)', prop: 'Allele_frequency_EAS_ExAC', core: true},
                     {title: 'FIN Allele frequency (ExAC minus TCGA)', prop: 'Allele_frequency_FIN_ExAC', core: true},
                     {title: 'NFE Allele frequency (ExAC minus TCGA)', prop: 'Allele_frequency_NFE_ExAC', core: true},
-                    {title: 'OTH Allele frequency (ExAC minus TCGA)', prop: 'Allele_frequency_OTH_ExAC', core: true},
-                    {title: 'SAS Allele frequency (ExAC minus TCGA)', prop: 'Allele_frequency_SAS_ExAC', core: true}
+                    {title: 'SAS Allele frequency (ExAC minus TCGA)', prop: 'Allele_frequency_SAS_ExAC', core: true},
+                    {title: 'OTH Allele frequency (ExAC minus TCGA)', prop: 'Allele_frequency_OTH_ExAC', core: true}
```
zfisch: was this a requested change?

website/loaders/markdown-id-headers.js
```diff
@@ -4,21 +4,33 @@
...
-		tokens[idx].attrs.push(['id', slug]);
-
-		if (originalHeadingOpen) {
-			return originalHeadingOpen.apply(this, arguments);
-		} else {
-			return self.renderToken.apply(self, arguments);
-		}
-	};
+    var originalHeadingOpen = md.renderer.rules.heading_open;
+    var explicitIDRegex = /[ ]?\(\(([^)]+)\)\)/;
```
zfisch: could you provide a comment for this regex :)

website/js/components/VariantSubmitter.js
```diff
@@ -135,15 +136,19 @@ const VariantSubmitter = React.createClass({
...
+        const submitterRows = cols.map(({prop, title, value, noHelpLink}) => {
             const isEmptyValue = util.isEmptyField(value);
             const rowItem = util.getFormattedFieldByProp(prop, data);
 
             return (
                 <tr key={prop} className={ (isEmptyValue && this.props.hideEmptyItems) ? "variantfield-empty" : "" }>
                     {
-                        helpKey
-                            ? <KeyInline tableKey={title} onClick={(event) => this.props.showHelp(event, helpKey)}/>
+                        !noHelpLink
```
zfisch: minor, but why the double negative?

website/js/VariantTable.js
falquaddoomi: It wasn't a requested change, no. It was one of those cases where I was looking at the code, the help text didn't match up (it has "Other" at the end), and it didn't make sense to me to have "Southeast Asian" at the end either, so I quickly edited it. Is it more appropriate to remove it here and file it as an issue instead?

website/js/components/VariantSubmitter.js
falquaddoomi: Heh, yeah, that's definitely odd; nice catch. I've reversed the clauses and removed the double-negative. (For reference the field actually isn't used yet, but I should probably specify it for the fields for which we don't have help text yet...)

website/loaders/markdown-id-headers.js
falquaddoomi: Good call, added. I also removed the part where it matches the optional space at the start; spaces aren't visible, so it's not necessary to remove it.

website/js/components/VariantSubmitter.js
falquaddoomi: I should probably also add that functionality to every place we use KeyInline (the main details display component and the allele frequency tile) if I do use it, too.

falquaddoomi:
Instead of peppering the field definitions with 'noHelpLink: true' that we'd have to manually remove later anyway, I slightly tweaked the logic for KeyInline to not display a help link if there isn't a tooltip; the links will establish themselves automatically if an entry is added to the help. I left 'noHelpLink' in as a way to manually suppress rendering the help link should we need to, but it shouldn't be necessary now.

I've tested it pretty thoroughly locally, so I'd be fine with merging it without a lot of extra review.

PR #829, opened 2018-07-19
adding clinvar SCV_Version and DateSignificanceLastEvaluated fields

izcram:
For ClinVar reports, add the accession version and the date when the clinical significance was last evaluated


pipeline/data_merging/variant_merging.py
```diff
@@ -705,10 +714,11 @@ def repeat_merging(f_in, f_out):
                     if new_value == old_value:
```
falquaddoomi: I realize this isn't part of the PR, and I'm sure it works fine, but it looks a little funny to me. Why not write it as "if new_value != old_value: (contents of current else block here)"?

pipeline/data_merging/test_aggregate_reports.py
```diff
@@ -14,7 +14,74 @@
...
+        self.columns = ['Source', 'Gene_symbol_ENIGMA', 'Genomic_Coordinate',
```
falquaddoomi: 👍 for improved code readability. :)

pipeline/clinvar/clinvar.py
```diff
@@ -21,6 +21,20 @@ def textIfPresent(element, field):
     else:
         return(ff.text.encode('utf-8'))
 
+
+def processClinicalSignificanceElement(el, obj):
```
falquaddoomi: I can see the logic in this being a separate function since both `referenceAssertion` and `clinVarAssertion` need it, but this seems like it should be a method of the class since it just sets fields in the object you pass in. Perhaps `referenceAssertion` and `clinVarAssertion` could inherit from an `assertion` base class that includes this function as a method?

pipeline/data_merging/variant_merging.py
izcram: Yes, indeed, I was also stumbling over this... I tried to be smart and remove it, because it is in principle not necessary. Since we are building a sets of distinct values the `list(set(new_value + old_value))` line later on should already take care of this. However, this changes the order of the fields in general. Per se not a problem, but then it turns out that text fields containing ',' will get scrambled...

pipeline/clinvar/clinvar.py
izcram: Yes, that's a good point! For me,  `referenceAssertion` and `clinVarAssertion` are not too similar, despite their name, therefore I prefer this composition approach. But I'll be touching this code in a subsequent iteration and might take your suggestion up again.

PR #834, opened 2018-07-20
Help page refactor

strbean:
Needs feedback from @amycoffin


amycoffin:
I don't want to be the person to merge the PR, but it all looks good to me.

amycoffin:
I'm a bit of a file hoarder myself, so my intuition is that they aren't doing much harm there until we have it fully rolled out and up-to-function. Everyone is different on this, though. I would say we can delete it in 2 months or so, but it might be helpful to keep around.

strbean:
Good point Marc! I think I left some commented out code as well...

strbean:
Removed, I'll merge once the tests pass. @Amy, we can always dig back in the git history for the deleted files if we need 'em

Issue #837, opened 2018-07-25
Tooltips no longer display since PR #834


falquaddoomi:
PR #813 introduced hover-over tooltips; the content for those tooltips is scraped from the help markdown with some assumptions about where the help markdown is stored and how field help entries are formatted.

PR #834 introduced a new structure for the help documentation, including both content changes and changes to where the markdown is stored in the filesystem. The parsing code for extracting tooltips (predictably) hadn't been updated to reflect these changes, so the tooltips and help links ceased to display.

zfisch:
has this already been addressed?

falquaddoomi:
@zfisch Yes, the main issue was addressed in PR #838 and pushed to beta. I thought I'd fixed the issue with the help sections not opening in PR #839, but apparently it's still broken...I'll reopen the PR and see if I can fix it.

FYI, the problem I *thought* I'd fixed was caused by minification settings for HTML generated from markdown being different in production than they are in dev. Specifically, the minifier was removing quotes around attributes that didn't contain whitespace (e.g. id="xyz"), and the code that checked to see if a section should be open based on the fragment was looking for those quotes. I relaxed the check to also matching non-quoted id's, but apparently that wasn't enough.

falquaddoomi:
This should be fixed by PR #842. (I'd misplaced a quote in the regex, causing the match to fail when the quotes weren't present, i.e. in production.)

PR #838, opened 2018-07-25
Fixes tooltips not displaying after changes to help documentation

falquaddoomi:
Addresses issue #837.

Modifies parsing code to recursively descend into new help docs structure, looking for markdown. The scraper now looks for `<h5>` tags in addition to `<h4>` tags, and can look for an explicitly denoted `<span class="term_entry" />` tag instead of just `<li>`'s with nested `<h4>`'s (the previous term-definition format). Modified help content in cases where it wasn't possible to automatically parse out the terms.

I made a few small formatting changes, too, like adding newlines to break up large blocks of text. Since the tooltip uses only the first paragraph of the text, the change in `content/help/research/variant-nomenclature.md` to `Reference_Sequence` does affect what's shown in the tooltip for that field.

### DIFF
--- a/website/content/help/default/variant-details-fields.md
+++ b/website/content/help/default/variant-details-fields.md
@@ -8,8 +8,11 @@ The Expert Reviewed Data Portal contains an abridged view of the variant data in
 
 * #### Gene ((Gene_Symbol))
   The name of the gene on which the variant was found, as named by the HGNC. This will be either _BRCA1_ or _BRCA2_.
+
 * #### HGVS Nomenclature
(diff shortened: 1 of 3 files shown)


falquaddoomi:
Seems there's a small issue in production: when clicking a field name, the code in Help.js that determines if a panel should be expanded looks for id="fragment", but the minifier removes the quotes around the attributes, causing that check to fail and the panel to not be expanded. I've revised the code to look for both id="fragment" and id=fragment, which I'll be pushing shortly.

PR #839, opened 2018-07-30
Field names now properly open the panel they target when clicked

falquaddoomi:
Fixes an issue where panels weren't expanding if they were referenced in the fragment on production. This appeared to be due to the minifier stripping double quotes around attributes values, causing the check for id="fragment" in the panel to fail.

### DIFF
--- a/website/js/Help.js
+++ b/website/js/Help.js
@@ -69,15 +69,15 @@ const Help = React.createClass({
         }
     },
 
-    shouldBeExpanded(fragment, {name, id, contents, list}) {
+    shouldBeExpanded(fragment, fragRegex, {name, id, contents, list}) {
         let slug = id || slugify(name);
         if (slug === fragment) {
             return true;
         }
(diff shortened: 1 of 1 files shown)


zfisch:
@strbean @falquaddoomi @izcram are we able to merge this and push it to beta? The only concern with pushing to beta is if we have any new code on master that we don't want included in next week's release.

falquaddoomi:
@zfisch There doesn't appear to be anything newer than this on master, and I believe this commit comes immediately after deploying to beta and fixing a bug. It should be safe to merge and deploy it.

zfisch:
@falquaddoomi interestingly this doesn't seem to work on dev or beta, though it does work on my local machine. the builds were all successful. could you take a look at this?

the issue I'm seeing is that if I click on a field name on the variant detail page it no longer links to the correct section on the help page (all sections are collapsed).

PR #847, opened 2018-08-10
Rem mupit col selector

zfisch:
Hides the mupit data from column selection.


zfisch:
@falquaddoomi @strbean awaiting your review!

zfisch:
@falquaddoomi thanks! makes sense regarding the dummy flag, i only removed it because i didnt see it being used :)

PR #848, opened 2018-08-13
update docker docs to account for multiple dockers in the repo

zfisch:


zfisch:
@izcram are there any other changes that would be helpful here? e.g. reference to local UTA

we can ignore the circle failure during review/iteration (it's due to a failed download of a resource from hgwdev, fix will be up soon enough).

zfisch:
@izcram i just made a few more updates.

should we include the `--network` flag here? if so, could you add a note to explain why it's used and if it can be omitted?

also, is there information already available on setting up a local UTA instance that we could link to in the docs? if not, could you write a little something about getting it set up so a future dev can just read through these docs and run the docker without error?

Thanks!

izcram:
`--network host` seems necessary in order to connect to the uta container, which exposes a port on 50827, from within the pipeline container.

You could add a comment for starting the UTA container:
`docker run -dit --name uta_20170629 -p 50827:5432 biocommons/uta:uta_20170629`
More information can be found [there](https://github.com/biocommons/uta/#installing-uta-locally)

PR #851, opened 2018-08-15
Fixes #846, mupit tile obscured

### LINKED ISSUE #846: mupit tile is obscured by data
the mupit tile on the variant detail page is obscured by data

falquaddoomi:
With respect to fixing issue #846, this PR adds an `onLoad` handler to the MupitStructure component, which triggers a tile relayout when the image inside the component loads. (The image loading a little late was previously causing it to change height after the initial relayout completed, causing the tile to be truncated.) In order to supply a callback for this extra prop, the MupitStructure component is now explicitly instantiated instead of via VariantTable's `replace` field. Rows which either have a truthy `replace` field or truthy `dummy` field will now be considered non-empty, since Mupit no longer uses `replace`, but does have `dummy: true`.

This PR also overhauls the tile relayout code that was previously implemented via a pretty inefficient mechanism that required redrawing the entire contents of the `VariantDetail` component with each relayout request. The new mechanism now only calls Isotope's `arrange()` method rather than using `forceUpdate()` or any other React re-rendering method. This new mechanism is now invoked when tile collapse/expand animations *actually* end (instead of trying to guess when they'd end via a precisely-timed `setTimeout()` call). It is also invoked directly instead of via `forceUpdate()`, which had previously triggered the IsoGrid component's `componentDidUpdate()` method, indirectly causing a tile relayout.

There were a few minor code-style changes as well, e.g. switching `let` to `const` when possible and fixing inconsistent indentation.

### DIFF
--- a/website/js/index.js
+++ b/website/js/index.js
@@ -32,7 +32,8 @@ var moment = require('moment');
 // faisal: includes for masonry/isotope
 var Isotope = require('isotope-layout');
 require('isotope-packery');
-
+import TransitionEvents from 'react/lib/ReactTransitionEvents';
+import debounce from 'lodash/debounce';
 
 var brcaLogo = require('./img/BRCA-Exchange-tall-tranparent.png');
 var logos = require('./logos');
@@ -441,7 +442,7 @@ function isEmptyDiff(value) {
 const IsoGrid = React.createClass({
     displayName: 'IsoGrid',
 
-    // Wrapper to layout child elements passed in
+    // wraps contents in a masonry-managed container
     render: function () {
         const children = this.props.children;
         return (
@@ -451,23 +452,9 @@ const IsoGrid = React.createClass({
         );
     },
 
-    // When the DOM is rendered, let Masonry know what's changed
-    componentDidUpdate: function() {
-        if (this.masonry) {
-            var that = this;
-            setTimeout(() => {
-                // prevents improper layout of mupit tile, and potentially others
-                // 300ms is the time it takes for the tile to expand
-                that.masonry.reloadItems();
-                that.masonry.arrange();
-            }, 300);
-        }
-    },
-
-    // Set up Masonry
+    // create masonry object to manage children's positions
     componentDidMount: function() {
-        if(!this.masonry) {
-            // i suppose we're doing this so the children exist when we create it?
+        if (!this.masonry) {
             this.masonry = new Isotope('.isogrid', {
                 layoutMode: 'packery',
                 itemSelector: '.isogrid-item',
@@ -477,10 +464,14 @@ const IsoGrid = React.createClass({
                 }
             });
         }
-        else {
+    },
+
+    relayout: function(fullRefresh) {
+        if (fullRefresh) {
             this.masonry.reloadItems();
-            this.masonry.arrange();
         }
+
+        this.masonry.arrange();
     }
 });
 
@@ -498,7 +489,7 @@ var VariantDetail = React.createClass({
     componentWillMount: function () {
         backend.variant(this.props.params.id).subscribe(
             resp => {
-                return this.setState({data: resp.data, error: null});
+                this.setState({data: resp.data, error: null});
             },
             () => { this.setState({error: 'Problem connecting to server'}); }
         );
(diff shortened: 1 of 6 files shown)


website/js/index.js
```diff
@@ -506,20 +499,31 @@ var VariantDetail = React.createClass({
...
+
     },
-    onChildToggleMode: function() {
-        this.props.toggleMode();
-        // we need to reparse the tooltips if the mode changed
-        this.setState({
-            tooltips: parseTooltips(localStorage.getItem("research-mode") === 'true')
-        });
-        this.forceUpdate();
+    componentWillUpdate: function(nextProps) {
```
falquaddoomi: onChildToggleMode was removed because the only thing it did aside from relayout the tiles via a forced update was set the tooltips, and this call was being skipped by the new mode changer in the navigation bar anyway. Changing it to componentWillUpdate (for the state-changing parts) and componentDidUpdate (for the tile relayout, which just involves the DOM) ensures that we always catch and appropriately handle the mode change.

website/js/index.js
```diff
@@ -475,10 +464,14 @@ const IsoGrid = React.createClass({
...
-        else {
+    },
+
+    relayout: function(fullRefresh) {
+        if (fullRefresh) {
             this.masonry.reloadItems();
-            this.masonry.arrange();
         }
+
+        this.masonry.arrange();
```
zfisch: does reloadItems() encompass arrange() or are both necessary in the case of a fullRefresh?

website/js/index.js
zfisch: nice!

website/js/index.js
```diff
@@ -539,14 +545,26 @@ var VariantDetail = React.createClass({
             return false;
         }
     },
+    relayoutGrid: debounce(function(fullRefresh) {
+        if (this.isogrid) {
+            this.isogrid.relayout(fullRefresh);
+        }
+    }),
+    relayoutOnCollapsed: function(collapser) {
```
zfisch: 👍

website/js/index.js
```diff
@@ -559,15 +577,9 @@ var VariantDetail = React.createClass({
...
-        const me = this;
-
-        setTimeout(() => {
-            // this forces a re-render after a group has expanded/collapsed, fixing the layout
-            // note that 300ms just happens to be the duration of the expand/collapse animation
-            // it'd be better to run the re-layout whenever the animation ends
-            me.forceUpdate();
-        }, 300);
+        // find the actual collapsing DOM element so we can attach a handler to its transition-end event
+        const collapser = collapsingElemParent.parentElement.parentElement.getElementsByClassName("panel-collapse")[0];
```
zfisch: is this our best approach to finding the element? not a huge deal but looks fragile at a glance

website/js/index.js
falquaddoomi: It's necessary to call `reloadItems()` when children are added or removed from IsoGrid, which currently occurs when you change modes. `arrange()` just moves the existing children around, which is what we want in most cases when tiles change size.

More in depth, `arrange()` seems to maintain some bookkeeping information between calls for interpolating the positions of elements that `reloadItems()` clears, which is why I made calling it conditional. This shows up as a race condition when you interrupt a layout-in-progress with `reloadItems(); arrange();`-- the tiles end up incorrectly positioned due to, I assume, not being aware of the move in progress.

website/js/index.js
falquaddoomi: Agreed, it's definitely flimsy. What this is effectively doing is getting refs for all the <Panels> that I use without actually having to add that into each one of them (which I was concerned would be a lot of work to pass back from the new tile types), but I ended up writing the code to pass back refs for those tile types anyway later on...

So, good catch there. 👍I'll rewrite it to use proper refs to the Panels as it should. It's late here at the moment, but expect at least one more commit tomorrow that implements that.

zfisch: Sounds good! No hurry, whenever you're ready :)

falquaddoomi: FYI, I opted to have each panel that calls onChangeGroupVisibility pass a ref to the header, which is way cleaner than attempting to find it from the event target. It's included in the latest commit.

PR #854, opened 2018-08-16
changing URL resource in download_resources_files.sh and adding check…

izcram:
…sum verificiation

### DIFF
--- a/pipeline/download_resources_files.sh
+++ b/pipeline/download_resources_files.sh
@@ -10,16 +10,38 @@ echo "Downloading into $1 ...."
 
 cd $1
 
+URL_PREFIX="http://brcaexchange.org/backend/downloads/resources"
 
-wget http://hgwdev.soe.ucsc.edu/~cline/BRCA/resources/refseq_annotation.hg38.gp
-wget http://hgwdev.soe.ucsc.edu/~cline/BRCA/resources/refseq_annotation.hg19.gp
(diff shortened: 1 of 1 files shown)


pipeline/download_resources_files.sh
```diff
@@ -10,16 +10,38 @@ echo "Downloading into $1 ...."
...
+files[hg19.fa.gz]=c1a06a3122bb9b3747ceeefb902aecd3ea9b9008
+files[hg19ToHg38.over.chain.gz]=4769aaa7a5a7837c3947604207da346b1dd19a8f
+files[hg38.fa.gz]=d329507d63d32405664053785b2bfdc02b3e6b28
+files[refseq_annotation.hg18.gp]=21911f9cd02f00b82eeb6068b7ab737c4d89f896
+files[refseq_annotation.hg19.gp]=e12092220926fe112da26ffdc8188b76aef83231
+files[refseq_annotation.hg38.gp]=9fb29159e516d2f28c2ec0e41a338c85ac699528
+
+for file in "${!files[@]}"; do
+    hash=$(${HASH} $file | cut -f1 -d' ')
+    if [ "$hash" != "${files[$file]}" ]; then
```
zfisch: should this be `==` instead of `!=` ?

izcram: no, I think it's fine. In case the file already exists but the hash doesn't match it gets redownloaded

Issue #816, opened 2018-07-05
Solve ambiguity for protPrior = 0.02


amycoffin:
protPrior can = 0.02 for the following reasons: 
- Silent (synonymous) mutation (implemented in calcVarPriors.py)
- Missense variant outside key domain (implemented)
- In frame indel outside CI/key domain (not yet implemented) 

This ambiguity is most relevant to the protein-level estimation nested tile. 

Since there is a case where splicing prior can equal 0.02 to as well, outside splice consensus region, the ambiguity has a small chance of propagating to applicable prior (overall estimation) if applicable prior is not checked against the proteinPrior and splicingPrior.

falquaddoomi:
> Since there is a case where splicing prior can equal 0.02 to as well, outside splice consensus region, the ambiguity has a small chance of propagating to applicable prior (overall estimation) if applicable prior is not checked against the proteinPrior and splicingPrior.

This ambiguity should be handled by the text in the overall probability subtile, underneath the graphic. The text is "Applicable Probability is from X", where X is determined with the following logic:

(Note that `splicingPrior` is the max of `refDonorPrior`, `refAccPrior`, and `deNovoDonorPrior`.)

- If `proteinPrior` is greater than `splicingPrior`, X is "Protein-level Estimation"
- Otherwise, `deNovoDonorPrior` is compared to `refDonorPrior`:
  - If `deNovoDonorPrior` is greater than `refDonorPrior`, X is "Splicing-level Estimation (De Novo Donor Splice Site Creation)"
  - Otherwise, `refDonorPrior` and `refAccPrior` are compared:
     - If `refDonorPrior` is greater than `refAccPrior`, X is "Splicing-level Estimation (Donor Splice Site Impact)"
     - Otherwise (i.e., if `refDonorPrior` is less than or equal to `refAccPrior`), X is "Splicing-level Estimation (Acceptor Splice Site Impact)"

amycoffin:
Once issues #831 and #832 are addressed, we can implement a frameshift flag either using getVarType() or getVarConsequnces(), or perhaps a combination of the two.

amycoffin:
This issue will be more easily addressed once #831 and #832 are closed

amycoffin:
@falquaddoomi has this one ALSO been solved without updates to frameshift flag?

amycoffin:
#831 and #832 did not resolve as expected, and this ambiguity persists. I think our best options are as follows:

[...]

Issue #832, opened 2018-07-19
Update conditionals to check for values in updated varConsequences list 


amycoffin:
This issue assumes that Issue #831 is resolved. 

getVarConsequences is used in 6 conditional statements to populate data based on variant consequence terms. The following methods utilize these conditionals. Most, if not all, of these conditionals should be able to be altered to basically use an "if in list" format in lieu of an "if ==" format: 

getPriorProbSpliceRescueNonsenseSNS (line 1401)
current:  `if getVarConsequences(variant) == "stop_gained" and varInExon(variant) == True: `
becomes: `if "stop_gained" in getVarConsequences(variant) and varInExon(variant ==True:
`

getPriorProbAfterGreyZoneSNS (1719),
getPriorProbSpliceDonorSNS (2216),
getPriorProbSpliceDonorSNS(2404), 
getPriorProbInExonSNS (2699), 
getPriorProbInUTRSNS( 3174)).

If the code is too buggy after changing these conditionals, revert getVarConsequences() and the associated methods above to a previous working version. Then, write an updated getVarConsequences as a new, isolated method that works and appends the column. We can problem-solve from there on how to get conditionals more accurate.

amycoffin:
@rcurrie I am assigning you to these issues for now, because you where there but if you don't feel comfortable implementing them, feel free to just unassign yourself (and assign me). We can figure out whether Melissa or Zack will. This loop will close somewhat easily.

rcurrie:
@amycoffin I can implement the 'in' logic but will need someone's cursory eyes on the dangling else clauses to verify that they can still be a catch all. There are a ton of consequences that can be returned and its not clear to me that all of them can be caught via the else.

amycoffin:
@rcurrie 
1. I think it is already unclear if or how elses are actually acting as a catch all, even prior to the if-in logic being implemented. 

2. the main changes we would expect to see after if-in logic is implemented is where the first if gets thrown more often. To generalize, say that right now 2 variants have varConsequences = consequence A. Implementing the Consequence list assignment to variants might mean that we now have 4 variants that have Consequence A in consequenceList. (Assuming that 2 additional variants had it as the second item in varConsequenceList and it got missed)
Assume that a function originally runs `if x == consequence A, then True Else, False` and then is updated to run `if consequence A in list, then True Else, false ` . It would return True 2 times in the first case, while showing True 4 times in the updated case. These are the types of changes I expect to see after the logic is implemented. 

[...]

Issue #831, opened 2018-07-19
Update getVarConsequences() methods to better reflect varConsequences


amycoffin:
**Current Functionality:**

getVarConsequences (line 247) currently calls Ensembl VEP to retrieve characteristics of a variant that are relevant to pathogenicity, also called "consequences." VEP returns a dictionary with the following structure, with bolded keys/values returned by current getVarConsequences().  

{'allele_string',..., **'transcript_consequences':** [{'amino acids': [], ..., **'consequence_terms': ['missense_variant,** splice_region_variant]}]}

Lines 274-281 indicate that only the first item in the consequence terms array is returned:
` return gene["consequence_terms"][0] `(line 281).

Referring to the above example, the current code dictates that 'missense_variant' is all that is returned by getVarConsequences, and we omit useful information, such as  'splice_region_variant' . 

**Action to Take**:
**Update getVarConsequences to return the entire consequence_terms array. Append this array to the end of the priors output as its own column, calling the column varConsequences**

getVarConsequences is used in conditionals to populate output data based on variant consequence terms (lines 1401, 1719, 2216, 2404, 2699, and 3174). Conditionals will need to be altered to address arrays rather than strings, but this will be handled in issue #832 .   

All possible consequence terms can be found [here](https://uswest.ensembl.org/info/genome/variation/prediction/predicted_data.html#consequence_type_table).

amycoffin:
@rcurrie I am assigning you to these issues for now, because you where there  but if you don't feel comfortable implementing them, feel free to just unassign yourself and we can figure out whether Melissa or Zack will. This loop will close somewhat easily.

rcurrie:
After talking with @amycoffin and indirectly @melissacline here's the plan:

- [ ] Modify getVarConsequences() to return a list
- [ ] Change all comparisons of the result to 'if in'
- [ ] Run the test 88 variants and see if the output is the same

Amy is going to look at the if's to see if they really are all mutually exclusive as well as the default else to see if it makes sense given there are 20+ consequences it implicitly resolves to.

rcurrie:
@melissacline @amycoffin Good news - Chloe's test 88 variants generated identical priors output! I'm going to do a bit more testing to see if any of the dangling else clauses are being taken or not - they may never occur. Will also see about stepping up to the full input just to see if it changes or not.

rcurrie:
[...]

PR #864, opened 2018-08-24
Adds ISPP tile and backend support

falquaddoomi:
This PR introduces a new tile, the In-Silico Probability of Pathogenicity (ISPP) tile, which displays data from a new model, the InSilicoPriors model (table name `insilicopriors`). The table is populated by splitting off the ISPP-specific columns from `built_with_priors.tsv` during variant processing in the `addrelease` management command, then inserting the ISPP entry after each variant that's inserted.

It also adds a research-mode help section for the ISPP tile and an extended "about" document at  `/about/insilicoScoring` (linked from the help section) on how probabilities are computed. As far as I know the text is still a WIP, so I've left in placeholders (e.g., "[Caption]") from the original documents.

### Backend Notes
- `deployment/deploy-data` has been modified to look for `built_with_priors.tsv` as its input in the release bundle; it's possible that that's not the correct filename.
- There isn't currently anything in the diffs about adding ISPP data; it'd be helpful to see that in the "previous versions" table IMHO.
- Currently the backend returns the priors data by adding a new key, `priors`, to the response from `/variants/?variant_id=x`, which contains the corresponding entry from the InSilicoPriors model. I'm not 100% sure how this interacts with variant versioning, but it seems to be working. <sup>[1]</sup>
- I added a library, `tqdm`, for rendering progress bars and estimating time remaining in the `addrelease` and `add_diff_json` management commands. It's a small, efficient library with no other prerequisites so I figured it was a modest addition, but it's easy to remove if we don't want it.

### Frontend Notes
- There are still two ambiguous probabilities in the UI, specifically Splicing-level De Novo Donor probability 0.02, and Protein-level probability 0.02. These may be resolved by some new flags already added or being added to the pipeline, but I think I'll need some biologist input either way.
- We should probably review the text that's displayed when we have missing probabilities. For instance, I've noticed that we're occasionally missing protein-level probabilities in an exonic variant (e.g., chr13:g.32316449:T>C), yet the message states the probability is missing "[...] because the variant is not inside an exon, and thus is not associated with estimated protein-level impact." Incidentally, that variant does have splicing-level data; it just appears to be missing the protein-level data despite being exonic.
  - For cases where none of the probabilities are present, I've opted to display a single "no data found" message instead of individually displaying the messages for each missing probability (which incorrectly suggest that the variant isn't exonic, for instance). We can revisit this later.

*[1] From my limited testing, the priors data appears to associate with the variant corresponding to the release in which the priors data was introduced. For example, in the array of variants returned from `/variant/?variant_id=X`, variant X only has a full 'priors' key for release 12 in which the priors data was imported, not in its versions from previous releases.*


website/django/data/models.py
```diff
@@ -478,3 +478,101 @@ class CurrentVariant(models.Model):
...
         db_table = 'currentvariant'
         managed = False
+
+
+class InSilicoPriors(models.Model):
+    Variant = models.OneToOneField('Variant', primary_key=True)
+
+    varType = models.TextField()
+    varLoc = models.TextField()
+    applicablePrior = models.TextField(null=True)
```
zfisch: previously we've avoided using null=True on text fields per recommendation from django docs https://docs.djangoproject.com/en/dev/ref/models/fields/#null

I'm open to being convinced otherwise though if there's a good reason for it :)

website/js/components/collapsible/CollapsibleSection.js
```diff
@@ -0,0 +1,81 @@
+/*eslint-env browser */
+/*global require: false, module */
+'use strict';
+
+import React from "react";
+import {CollapsableMixin} from "react-bootstrap";
+import classNames from 'classnames';
+// import util from '../../util';
```
zfisch: can these comments be removed?

website/js/components/insilicopred/InSilicoPredSubtile.js
```diff
@@ -0,0 +1,114 @@
...
+}
+
+
+export default class InSilicoPredSubtile extends React.Component {
+    render() {
+        const {probability, reason, varLoc, varType} = this.props;
+        const summarizedData = {varLoc, varType};
+        const cols = [
+            { title: 'VEP Consequences', prop: 'varLoc' },
+            // { title: 'Variant Type', prop: 'varType' }
```
zfisch: can be removed?

website/js/components/insilicopred/ProteinLevelSubtile.js
```diff
@@ -0,0 +1,125 @@
...
+    nonsenseGrayZonVal: 0.5,
+    nonsenseVal: 0.99
+};
+
+function labelClasses(v) {
+    return classNames('label-field borderless ', mapValToColor(v));
+}
+
+function rowClasses(v, probability) {
+    // FIXME: needs logic to disambiguate cases where the probability matches multiple rows
```
zfisch: is this a must fix now or fix at some point kind of situation?

website/js/components/insilicopred/SilicoPredTile.js
```diff
@@ -0,0 +1,167 @@
...
+            variant: {
+                sequence: data.altRefAccSeq,
+                MES: data.altRefAccMES,
+                zScore: data.altRefAccZ
+            }
+        }
+    };
+}
+
+function numberify(x) {
```
zfisch: 👍

zfisch: this function name makes me happy

website/js/components/insilicopred/SilicoPredTile.js
```diff
@@ -0,0 +1,167 @@
...
+            },
+            closest: {
+                sequence: data.closestDonorRefSeq,
+                MES: data.closestDonorRefMES,
+                zScore: data.closestDonorRefZ,
+                donorPosition: useTranscriptSplicePos ? data.closestDonorTranscriptSplicePos : data.closestDonorGenomicSplicePos
+            },
+            closestAlt: {
+                sequence: data.closestDonorAltSeq,
+                MES: data.closestDonorAltMES, // where is this?
```
zfisch: ?

website/js/components/insilicopred/SplicingLevelSubtile.js
```diff
@@ -0,0 +1,455 @@
...
+            },
+            {
+                check: () => true, // the default option, if reached
+                text: (<span>If this variant creates a de novo donor, the de novo donor is not predicted to correct disruptions to the translation frame.</span>)
+            }
+        ]
+    },
+];
+
+function getPathosLevel(prob, isDeNovo, data) {
```
zfisch: such a sad double entendre...

website/js/components/insilicopred/SplicingLevelSubtile.js
```diff
@@ -0,0 +1,455 @@
...
+
+            <table className="splicing-level-table">
+                <tr>
+                    <td width="60%">Qualitative Category</td>
+                    <td width="30%">Z-score Range</td>
+                    <td width="20%">Probability</td>
+                </tr>
+
+                {
+                    // FIXME: highlight row that matches the prior prob value
```
zfisch: now or later?

website/js/components/insilicopred/SplicingLevelSubtile.js
```diff
@@ -0,0 +1,455 @@
...
```
zfisch: same

website/js/components/insilicopred/SplicingLevelSubtile.js
falquaddoomi: Forgot to remove that; it's been implemented two lines below, where it checks if the row's probability matches the prior prob., and also validates an optional check condition.

Expanding on the "check", there is one case, prob == 0.04, which has two matching rows; the check function defined for those rows disambiguates which one should be highlighted.

website/js/components/insilicopred/SplicingLevelSubtile.js
falquaddoomi: Same as the above. In this case, though, we're unfortunately still missing the condition for prob == 0.02. I've added a FIXME to that collection indicating where the check conditions should go.

website/js/components/insilicopred/SilicoPredTile.js
falquaddoomi: Ah, that's a cruft comment. The field was mentioned in the wireframes, but wasn't included in Amy's initial version of the document listing the ISPP fields; it's since been added, and it's in the output.

website/js/components/insilicopred/SilicoPredTile.js
falquaddoomi: 😄

website/js/components/insilicopred/ProteinLevelSubtile.js
falquaddoomi: We're still missing the condition that disambiguates the two possible matches for protein-level prob == 0.02, documented in issue #816. This issue's apparently waiting on #831 and #832 before it can be resolved.

Visually this will currently appear as both rows being highlighted. I don't know whether that's acceptable for letting people take a look at the data, or if we should hold off until we've resolved #816.

deployment/deploy-data
```diff
@@ -21,7 +21,8 @@ function add_release_tar_to_db {
 
     RELEASE_DIR="${STAGING}/output/release"
 
-    DATA=${RELEASE_DIR}/built_with_change_types.tsv
+    # DATA=${RELEASE_DIR}/built_with_change_types.tsv
```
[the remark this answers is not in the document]
falquaddoomi: Ah, good to know. I'll revert to the original, then.

website/django/data/models.py
falquaddoomi: Sounds good to me; I agree that we should avoid having multiple values that represent null.  (I will say the nice thing about using null to represent an empty value is that it's available irrespective of type: any column type can all be nullable, whereas a special string value for null strings means you need to treat other possibly empty types differently.)

Anyway, Should those fields instead have `default=''` or `default='-'` specified, then? I've noticed that both are used to represent empty values and wasn't sure which one is preferred (I'd personally prefer '' if we're going to use a string).

website/js/components/insilicopred/InSilicoPredSubtile.js
falquaddoomi: Good catch, both those sets of comments are unnecessary.

website/django/data/migrations/0035_merge.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
falquaddoomi: I agree that the apparently duplicated migration is confusing -- it's because this branch was originally based on master before Marc added that migration. The "empty" migration is a merge migration, created via `./manage.py makemigrations --merge`, which reconciles the two colliding migrations. It passed the tests because it's technically correct, but in retrospect pretty confusing.

Anyway, I think the better thing to do would be to remove both migrations that I added, rebase from master so that I'm sure all the migrations are in, and then regenerate it as a single, new migration so that we have a linear history.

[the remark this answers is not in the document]
falquaddoomi: Ok, I've reverted the migration history to 0032, applied 0033 (Marc's migration), and re-added a single migration for insilicopriors (0034) that keeps the migration history linear.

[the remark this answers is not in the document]
zfisch: gotcha, I figured this was probably a django thing but I think your adjustments are much more dev friendly, thanks!

PR #878, opened 2018-09-05
Revert "Revert "Adds ISPP tile and backend support""

falquaddoomi:
So, turns out that my reversion of my PR was a little premature; it turned out to be out-of-sync database state that caused the migration to fail (since it had effectively already been applied). Anyway, this PR reverts my reversion, bringing back the contents of PR #864 prior to the reversion.

From Git: This reverts commit abec4f14ae529bfc871990a65ce78ccdb143121c.


zfisch:
here goes!

PR #884, opened 2018-09-13
Feature/streamlining pipeline running

izcram:
A initial PR for the 'Makefile' approach to run the pipeline.

It focuses on easily generating a new data release. 
In a nutshell we should be able to do this using the following command on the pipeline machine:
`/home/pipeline/brca_upstream/pipeline/pipeline_running/generate_release.sh /home/pipeline/monthly_releases /home/pipeline/luigi_pipeline_credentials.cfg /home/pipeline/previous_releases`

Support for pipeline development work will be addressed in a subsequent PR. Note that the generation of a brca_pipeline_cfg.mk outside the context of data releases will come soon in that context.

Some details still need to be fixed and tested, but it is ready enough for a review.


pipeline/pipeline_running/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+TODO: comming soon, additional instructions from Mac OS X
+
+## Creating a New Data Release
+
+New data releases should ideally be generated on a dedicated pipeline machine. Although it could be run on any other machine in principle, some sources (e.g. LOVD) are only available from the pipeline machine.
+
+### Credentials
+
+Some early stages of the pipeline new some credentials to download data. These can be passed into the container by mounting an appropriate file. Also note, that some data sets are only available via the pipeline machine. However, later stages of the pipeline don't need any and some dummy file could be created.
```
rcurrie: delete 'some'

rcurrie: Delete 'some'

pipeline/pipeline_running/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+Currently, that a credential files should contain the following:
```
rcurrie: Where and what name is this credentials file? docker-compose defaults to looking in .env and I usually include a template, sample_env that is checked into git and make sure .env is prevented from being commited in .gitignore. See https://github.com/UCSC-Treehouse/hub as a sample

pipeline/Makefile
```diff
@@ -0,0 +1,109 @@
...
+
+checkout: ## checking out code base
+	[ -d $(CODE_BASE) ] || git clone $(BRCA_GIT_REPO) $(CODE_BASE)
+	cd $(CODE_BASE) && git checkout $(GIT_COMMIT)
+
+build-docker: ## build main pipeline docker image
+	cd $(CODE_BASE)/pipeline/docker && ./build_docker_image.sh $(GIT_COMMIT) $(RELEASE_TAG) || { echo "$(date) Something wrong with docker image creation"; exit 1; }
+
+download-resources: ## download resources files
+	$(CODE_BASE)/pipeline/download_resources_files.sh $(RESOURCES_DIR) >> "$(LOG_DIR)/download_resources_$(RELEASE_TAG).log" 2>&1 || { echo "Downloading resource files failed"; exit 1; }
```
falquaddoomi: `download_resource_files.sh` removes the .gz files when they're unzipped, which causes them to be re-downloaded with each run. I realize this is a problem with that script and not this PR per se, but it's something to think about.

pipeline/Makefile
```diff
@@ -0,0 +1,109 @@
...
+
+.ONESHELL:
+setup-lovd: ## setting up LOVD data via pipeline machine, if not running on pipeline machine, as the access is IP restricted. Requires access to main pipeline machine	
+	if [ `hostname -s` != "brca-pipeline" ]; then
+		mkdir -p $(OUT_DIR)/LOVD
+		ssh pipeline@52.151.62.136 curl https://databases.lovd.nl/shared/export/BRCA > $(OUT_DIR)/LOVD/BRCA.txt
+	fi
+
+# TODO: docker access within docker doesn't work yet on Mac OS X
+DOCKER_RUN_CMD = docker run -it --rm -u `id -u ${USER}`:`stat -c '%g' /var/run/docker.sock` \
```
falquaddoomi: Minor note, but if we plan to support OS X, the formatting flag for `stat` is `-f` instead of `-c`.

pipeline/Makefile
```diff
@@ -0,0 +1,109 @@
...
+	cd $(CODE_BASE)/pipeline/docker && ./build_docker_image.sh $(GIT_COMMIT) $(RELEASE_TAG) || { echo "$(date) Something wrong with docker image creation"; exit 1; }
+
+download-resources: ## download resources files
+	$(CODE_BASE)/pipeline/download_resources_files.sh $(RESOURCES_DIR) >> "$(LOG_DIR)/download_resources_$(RELEASE_TAG).log" 2>&1 || { echo "Downloading resource files failed"; exit 1; }
+
+
+start-local-uta: ## starting local uta docker container
+	[ `docker ps -f name=$(UTA_CONTAINER) | wc -l` -gt 1 ] || docker run -dit --name $(UTA_CONTAINER) -p $(UTA_PORT):5432 $(UTA_DOCKER_IMAGE)
+
+.ONESHELL:
```
falquaddoomi: Another OS X note: the default version of `make` shipped with OS X (at least on my version, 10.13.6) doesn't support `.ONESHELL`. Upgrading to 3.82 fixes the problem.

pipeline/Makefile
rcurrie: 2 cents: Trying to support two platforms has a cost. I go back and forth with this (spent about 2 hours last night fighting with OSX so I can run tensorflow while on a plane, which requires python 3.5, which causes pain on OSX....poof...2 hours. I haven't tried it on a Mac but in theory if the docker daemon is run in the first linux container (russian doll style) it might work. That said ultimately I think we should try and get rid of dind. In other projects I work on it can cause problems due to host daemon incompatibility issues - sometimes subtle. Ideally we have a fairly basic parent script that then calls a series of dockerized ingest and transform containers.

izcram: Yes, it has a cost. However, it seems not an excessive one: instead of using `stat` we can just use a linux compatible `gstat` version on mac, same with `make` and `gmake`. All a dev has to do is to brew install these. The docker-in-docker I still have to figure out. But I don't consider this to be part of this PR, because here we only care about generating new data releases, which happen on Linux machines.

pipeline/Makefile
izcram: yes, good point! will fix it.

izcram:
@zfisch If there are no major points, feel free to merge it. I've setup a crontab on the pipeline machine:

0 6 * * 6 source /home/pipeline/.profile; cd ${HOME}/brca_upstream/; git checkout master; git pull origin master; ${HOME}/brca_upstream/pipeline/pipeline_running/generate_release.sh /home/pipeline/monthly_releases /home/pipeline/luigi_pipeline_credentials.cfg /home/pipeline/previous_releases

So, if things go well, we should have a fresh data release generated on Saturday

pipeline/pipeline_running/generate_release.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+ROOT_DIR=$(realpath $1)
+CREDENTIALS_PATH=$(realpath $2)
+PREVIOUS_RELEASE_DIR=$(realpath $3)
+
+GIT_COMMIT=${4-:"master"}
```
zfisch: I believe this should be GIT_COMMIT=${4:-"master"}

izcram: yes, good catch!

PR #892, opened 2018-09-20
update HCI links to be uniform

zfisch:
This PR updates links to HCI as requested by Chunwei at Sean's lab.

@falquaddoomi, I'm noticing a bunch of other non-changes that are showing diffs (as well as automatically removed whitespaces) -- is this a peculiarity of the markdown setup we're using?


falquaddoomi:
Regarding the whitespace differences, I assume those were somehow introduced by your editor. The markdown files are never written to by the site, at least.

zfisch:
@falquaddoomi yep, you're right -- I have sublime text set up to trim whitespace -- turns out the lines that don't show trimmed whitespace do show if I run `git diff --ws-error-highlight=all`.

PR #894, opened 2018-09-21
Fixes empty transcript viz. legend boxes on Firefox

falquaddoomi:
Firefox doesn't allow SVG elements' widths and heights to be specified via CSS, unlike most other browsers. This causes the legend boxes to display with width and height 0, rendering as an unfilled box.

The fix is simply to specify the width and height as properties on the `<rect>` elements.


falquaddoomi:
That's a good idea. I don't know what our priorities are for browser support, but maybe reviewing the access logs to see what people are using would be a good start.

I can also resume our conversation with BrowserStack on getting a free account for their browser-testing services. I'm thinking we can start with manually reviewing screenshots of the website in our supported browsers, then think about how to automate that process later on. On a separate note, I've been looking into automated front-end testing, but that wouldn't catch these kinds of browser inconsistencies.

PR #897, opened 2018-09-24
Feature/better local dev support

izcram:
* generating docker image out of current code instead of code in a remote repo
* new target `init`, which sets up brca_pipeline_cfg.mk
* make file targets should also run on OS X now
* new target `run-task`: can run arbitrary Luigi task, e.g. `make run-task ConcatenateESPBRCA12Data`

Two more features will be added in a separate PR later:
* force run luigi task (i.e. deleting appropriate files before running `run-task`)
* "visualize" luigi graph (perhaps on the command line only)


pipeline/luigi/CompileVCFFiles.py
```diff
@@ -141,23 +141,46 @@ def extract_file(archive_path, tmp_dir, file_path):
 utilities_method_dir = os.path.abspath('../utilities')
 
 
-###############################################
-#                   CLINVAR                   #
-###############################################
-
-
-class DownloadLatestClinvarData(luigi.Task):
+class BRCATask(luigi.Task):
```
zfisch: 🙌

pipeline/docker/Dockerfile
```diff
@@ -19,13 +19,13 @@ RUN curl -fsSL get.docker.com -o get-docker.sh \
...
 WORKDIR /opt
 
-COPY requirements_docker.txt .
-COPY test-requirements_docker.txt .
+COPY pipeline/requirements.txt .
+COPY test-requirements.txt .
 
 # install numpy first to avoid issues with bio python and bx-python (see also https://github.com/LUMC/vep2lovd/issues/1)
-RUN pip install $(grep numpy requirements_docker.txt)
+RUN pip install $(grep numpy requirements.txt)
```
rcurrie: Nice!

.dockerignore
```diff
@@ -0,0 +1,5 @@
+*.img
```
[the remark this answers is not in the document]
izcram: yes, good point!
img is actually not necessary, was more a relict from where I copied stuff from.

PR #907, opened 2018-10-10
handle priors for deleted variants

zfisch:


zfisch:
Quick Synopsis:

During the diff process, we identify and store all deleted variants in a file (removed.tsv) by checking for all of the variants from a previous release in the current release data https://github.com/BRCAChallenge/brca-exchange/blob/master/pipeline/utilities/releaseDiff.py#L681

When adding the new release data to the db, we add new versions of all deleted variants from removed.tsv to the database with a change_type of "deleted" https://github.com/BRCAChallenge/brca-exchange/blob/master/website/django/data/management/commands/addrelease.py#L112.

This allows us to show which variants were deleted to our users without changing anything about previous versions of those variants.

PR #914, opened 2018-10-19
fix gene symbol association with variants in clinvar source file

zfisch:
Handles cases where multiple MeasureRelationship attributes are present with different Gene Symbol values. Should lead to an increased number of ClinVar variants and fewer incorrectly removed variants in the future.

### DIFF
--- a/pipeline/clinvar/clinVarBrca.py
+++ b/pipeline/clinvar/clinVarBrca.py
@@ -30,9 +30,10 @@ def main():
     parser.add_argument('-a', "--artifacts_dir", help='Artifacts directory with pipeline artifact files.')
     args = parser.parse_args()
 
-    logging_level = logging.DEBUG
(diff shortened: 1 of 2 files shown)


zfisch:
@izcram I'm seeing now that you've made a similar update in https://github.com/izcram/brca-exchange/commit/120c0a9762b9513f1b2262d5cd1f9ae841bdd24c, which would explain why we're seeing those variants again in your data -- feel free to close this pr if you'd like.

izcram:
your fix is less invasive than mine but I took the opportunity to speed up the xml filtering significantly https://github.com/BRCAChallenge/brca-exchange/pull/915/files#diff-8eaa78d469d3157e0fa0fdbc1131e95d since the latter PR should be merged soon, I propose to incorporate that change.

PR #920, opened 2018-10-25
allow searches for complete clinvar accession numbers

### LINKED ISSUE #918: add clinvar accession search key

zfisch:
Resolves #918 

Mostly for internal purposes currently, but could be expanded if users want. Only works if a complete clinvar accession number is provided. Doesn't handle version numbers.

E.G. "SCV000053796" would return a result but "SCV000053796.2" wouldn't.


zfisch:
Opted to update and accept version numbers in the ClinVarAccession_ENIGMA field.

izcram:
LGTM! thanks, that should be useful!

PR #928, opened 2018-11-04
Splicing Priors: performance improvements and refactored code

falquaddoomi:
This (work-in-progress) PR includes a major refactoring of the splicing priors code and a few performance improvements. This is kind of a "checkpoint" PR while I switch temporarily to another project; I intend the PR to be a place where people can leave commentary after looking through or testing the branch.

The refactor splits the contents of `pipeline/splicing/calcVarPriors.py` into a new package, `calc_priors`, with multiple submodules roughly described below:

- `constants`: constant values mostly from the first part of `calcVarPriors.py` but also from other locations in the code.
- `dataproc`: contains the definition of an empty splicing priors object, and methods for reading in variants from the input and writing variants plus priors data to the output.
- `verify`: contains predicates for verifying different aspects of the input variants. (Some non-predicates remain due to circular dependencies, but these should be moved eventually.)
- `extract`: contains methods that extract and transform variant data into values used for the probability assessments later on.
- `compute`: this module is somewhat the same as `extract`, but was split into a separate module to remove many of the circular dependences between `extract` and `verify`.
- `priors`: this is where the "heavy lifting" of assessing and returning probabilities occurs.
- `utils`: Contains a few methods for supporting the integration tests (i.e., benchmarking, comparing TSVs cell-by-cell, and validating the approximate equivalence of floating-point values).

The performance improvements come primarily from switchting to [maxentpy](https://github.com/kepbod/maxentpy) for computing MES scores instead of launching a Perl interpreter in a new process. A few places where static files were loaded per-variant have been factored out into a single initial loading or inlined as constants.

We can probably remove the perl code and the tests comparing it to maxentpy once maxentpy's output is sufficiently vetted.


pipeline/splicing/calc_priors/dataproc.py
```diff
@@ -0,0 +1,144 @@
...
+        return self._data[key]
+
+    def __len__(self):
+        return len(self._data)
+
+    def __iter__(self):
+        return iter(self._data)
+
+
+# FIXME: make this read-only again and update addVarDataToRow to base its copy off this
```
zfisch: is this a must fix or can it be addressed in a followup? i.e. will it cause errors?

pipeline/splicing/calc_priors/dataproc.py
```diff
@@ -0,0 +1,144 @@
...
+
+    varDict = {"varGene": variant["Gene_Symbol"],
+               "varChrom": variant["Chr"],
+               "varStrand": varStrand,
+               "varGenCoordinate": variant["Pos"],
+               "varType": varType,
+               "varLoc": varLoc,
+               "varHGVScDNA": variant["HGVS_cDNA"]}
+
+    return varDict
```
zfisch: add nl please

pipeline/splicing/calc_priors/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    :param a:
+    :param b:
+    :param rel_tol: the relative tolerance
+    :param abs_tol: the minimum absolute tolerance
+    :return: True if the values a and b are close to each other and False otherwise.
+    """
+    return abs(a - b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
+
+
+class MismatchException(Exception):
```
zfisch: ?

zfisch:
Also, could you coordinate with @amycoffin to remove variants from blacklisting that no longer need it?

pipeline/splicing/calc_priors/dataproc.py
falquaddoomi: It's a "nice-to-have"; it'll work as it is.

Now that I'm thinking about this again, rather than worry about making the dict read-only it'd be better to add a class to represent what we're currently piling into BLANK_DICT, but that can be done in a followup.

zfisch:
Hit the following error rerunning an old release through this updated splicing pipeline:
```
Traceback (most recent call last):
  File "calcVarPriors.py", line 3560, in <module>
    cli()
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "calcVarPriors.py", line 3556, in calc
    ctx.obj["genome"], ctx.obj["transcripts"], ctx.obj["processes"])
  File "calcVarPriors.py", line 3493, in calc_all
    calculatedVariants = pool.map(calc_one, list(inputData))
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 251, in map
    return self.map_async(func, iterable, chunksize).get()
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 567, in get
    raise self._value
requests.exceptions.ConnectionError: None: Max retries exceeded with url: /cgi-bin/das/hg38/dna?segment=chr13:32394789,32394805 (Caused by None)
zfisch@zfisch:~/splicing/brca-exchange/pipeline/splicing$ docker run -it --rm --user=`id -u`:`id -g` -v /home/zfisch/splicing/references:/references:ro -v /home/zfisch/splicing/data:/data brcachallenge/splicing-pipeline calc /data/built_with_mupit.tsv /data/built_with_priors.tsv
```

Not sure if it's particular to this branch, but may as well address it here since it will nuke pipeline runs -- could you tackle this @falquaddoomi?

zfisch:
@falquaddoomi rerun produced output without error though it may be worth reconsidering the rerun logic. There are quite a few changes as well as a missing varConsequences field in the new output. Could you take a look and share your thoughts on whether these changes are intentional?

Here's a link to the original built_with_priors.tsv from the 11/3/18 release as well as an updated output using the new pipeline: https://drive.google.com/file/d/1H2tLQ0xIqvai6jug9hgrWxBGg6rMVEzH/view?usp=sharing

zfisch:
Hi @falquaddoomi and happy new year! Last we spoke I remember you had just fixed a bug herein and confirmed matching outputs from a given input. Anything else you'd like reviewed or tweeked? If not, feel free to update and merge whenever you're ready :)

zfisch:
@falquaddoomi bump ☝️

PR #963, opened 2019-01-03
don't display literature results or make request for data on summary …

zfisch:
…view

### DIFF
--- a/website/js/index.js
+++ b/website/js/index.js
@@ -1114,9 +1114,12 @@ var VariantDetail = React.createClass({
                 </Row>
 
                 <Row>
-                    <Col md={12} className="variant-literature-col">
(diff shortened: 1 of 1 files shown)


website/js/index.js
```diff
@@ -1114,9 +1114,12 @@ var VariantDetail = React.createClass({
                 </Row>
 
                 <Row>
-                    <Col md={12} className="variant-literature-col">
-                        <LiteratureTable maxRows={10} variant={variant} hideEmptyItems={this.state.hideEmptyItems} />
-                    </Col>
+                    { this.props.mode === "research_mode" && (
```
falquaddoomi: I'd prefer to hide the `<Row>` element as well, just in case it takes up space even if there aren't any columns in it.

zfisch: good call, just updated

PR #975, opened 2019-01-21
Left-justifies allele frequency, LOVD report subtile headers

### LINKED ISSUE #971: Fix Indentations on Variant Details Tiles
The Allele Frequency Reference sets tile indents are not aligned, or rather, right aligned:
screen shot 2019-01-10 at 3 43 29 pm
I personally think that subtile titles (blue rows) should be Left aligned, to sort of mirror the main titles, but maybe this is just a matter of opinions and doesn't matter much. Right alignment makes sense for normal data labels (white rows).

This one is less of an issue, but I did notice that the LOVD tile's indentations seem unnecessarily far from the margin:
screen shot 2019-01-10 at 4 05 18 pm

Can we align the Allele Frequency tile labels, and perhaps move over the LOVD labels a bit?

falquaddoomi:
The allele frequency and LOVD report subtile headers are now left-aligned (instead of aligned with the other within-tile fields) and take up the full horizontal space within their tile. ClinVar reports are unchanged. Fixes #971.

Examples:
screenshot 2019-01-21 at 11 51 47

screenshot 2019-01-21 at 11 52 22

### DIFF
--- a/website/js/css/custom.css
+++ b/website/js/css/custom.css
@@ -1175,6 +1175,11 @@ div.variant-submitter-group .submitter-header td {
     flex: 0 0 203px; /* the +3 gets it to line up with the columns below it */
 }
 
+.lovd-submitter .submitter-header .submitter-cell.submitter-label {
+    text-align: left;
+    flex: none;
+}
+
 .allele-frequency-header td {
     border-bottom-width: 3px;
 }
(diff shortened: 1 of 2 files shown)


falquaddoomi:
@amycoffin Correct.

website/js/css/custom.css
```diff
@@ -1198,8 +1203,9 @@ div.variant-submitter-group .submitter-header td {
 .allele-frequency-header .allele-frequency-cell.allele-frequency-label {
     font-weight: bold;
     white-space: nowrap;
-    text-align: right;
-    flex: 0 0 203px; /* the +3 gets it to line up with the columns below it */
+    /*text-align: right;*/
```
zfisch: any reason to not delete these?

Issue #960, opened 2018-12-18
update data deployment


zfisch:
1. Need to be confident deployments will succeed and get a meaningful error if not.

2. Make sure deployments don't lock tables so users can still access (if they take > ~10 min)

3. Include simple commands to deploy locally, to dev, beta, and production. Really should be able to deploy to any machine but those should be automated.

4. Maintains existing functionality (deploys data, updates downloads directory with latest release, pushes to genome browser).

5. Explore options for coupling literature search and data deployment (currently separate scripts).

6. Can we get it to run in 5-15 minutes? Last deployment (12/6/18 data deployed to beta) took ~2 hours to complete after the first attempt crashed after ~8 hours.

zfisch:
@falquaddoomi @izcram either of you particularly interested (and available) to work on this? No major rush, but worth putting on the radar. I can work on it if not.

falquaddoomi:
Sure, I'd like to look into it.

zfisch:
@falquaddoomi Awesome! Let me know if you need anything along the way or have any questions. The master script is at https://github.com/BRCAChallenge/brca-exchange/blob/master/deployment/deploy-data.

falquaddoomi:
@zfisch Just to give a quick status update, I've been focusing almost completely on getting it to run quickly in a single transaction (which would prevent it from leaving the db in an intermediate state should it fail, at least). I've found that loading the variant/report diffs into a temporary table via `copy from`, then using an `insert-select` loop to associate them with their respective variants/reports is a lot faster than doing it in straight Python, at least on my local machine. I've made a few other minor cleanups and redundant code removals that make the code easier to work with, imho.

There are a few issues with the version of postgres that we're using (9.3) that'd be resolved by moving to a more contemporary version (e.g., 11). For instance, there appear to be some diffs that refer to the same variant/report, which violates the uniqueness constraint on the variant_id/report_id column in the diff table. Postgres 9.5 introduced the `on conflict do nothing` clause to `insert`, but we're unfortunately running 9.3. Also, indexing the `Genomic_Coordinate_hg38` column would improve performance quite a bit since we use it for looking up variants frequently, but since it can be quite long (more than 70,000 characters in some cases) it's too big for a regular btree index. Instead, we'd have to use a hash index, but hashes weren't great until they received attention in postgres 10.

zfisch:
[...]

PR #989, opened 2019-02-12
data deployment updates

falquaddoomi:
This PR (incompletely) addresses issue #960. I was intending to do some more vetting of the results, but I figure it's better for the code to be out for other people to test, too.

Anyway, this PR introduces a few changes:
- runs the entire data loading operation in a single transaction (previously, the add_diff_json command was being launched separately, causing it to run in a separate transaction).
- adds indices on a few fields that are used frequently in both loading data and querying:
  - **reports:** indexes SCV_ClinVar and Submission_ID_LOVD columns  separately, and adds a compound index of each of them together with the Data_Release column (since we often query for a data release *and* those columns).
  - **variants:** adds a hash index on the Genomic_Coordinate_hg38 column, which should improve performance both in loading and querying the database when serving pages. (We have to use a hash index instead of the more common and versatile btree index due to the occasionally huge values it takes on, e.g., 70,000+ characters in some cases. Wrt. versatility, hash indices only allow exact comparisons, but we're only using it for exact comparisons anyway. The performance improvement of using hash index wasn't great before postgres 10, but since we're running 10 on the server we should be fine.)
- uses postgres's COPY FROM to load report diffs into a temporary table, then moves the association of those report diffs with their respective reports into an "insert-select" query (rather than iterating in Python, which is slow). I'd like to do the same for the variants/in-silico priors data loading, but the logic is somewhat complicated and they don't take very long anyway (approx. 3-4 minutes versus the 10-15 for report diffs).
- minor refactors of the code to decrease code duplication and generally reduce verbosity.

Remaining questions:
- for some reason, there are four *extra* reports in the table after running this PR's implementation versus the original code (913,253 versus 913,249). I'm still trying to identify which four reports those are and why they're being included now, which I hope to wrap up this week.
- the rest of the data should be vetted further, but there are at least the same number of records in all the other relevant tables (variant, insilicopriors, and data_reportdiff). If anyone knows of a nice way to diff data between postgres database and/or dumpfiles, I'd appreciate it. :)


zfisch:
Played with this a little today, it truly is a major improvement in performance.

I actually got 6 more reports running the old code than I did with the code in this PR using the latest dataset on dev (/data/monthly_releases/data_release_2019-02-13/brca_out/release-02-13-19.tar.gz). 

I also noticed the disparity is only among LOVD reports:

```
zfischcopyone.pg=# SELECT "Source", COUNT ("Source") FROM report GROUP BY "Source";
    Source    | count
--------------+--------
 1000_Genomes |  21837
 BIC          | 515774
 ClinVar      | 496758
 ENIGMA       |  70726
 ESP          |  10149
 ExAC         |  36013
 exLOVD       |   5138
 LOVD         | 164894
(8 rows)

zfischcopytwo.pg=# SELECT "Source", COUNT ("Source") FROM report GROUP BY "Source";

    Source    | count
--------------+--------
 1000_Genomes |  21837
 BIC          | 515774
 ClinVar      | 496758
 ENIGMA       |  70726
 ESP          |  10149
 ExAC         |  36013
 exLOVD       |   5138
 LOVD         | 164900
(8 rows)
```

Didn't have enough time to get deeper than that but wanted to share findings.

website/django/data/utilities.py
```diff
@@ -164,3 +164,16 @@ def update_mupit_structure_for_existing_variants():
                 setattr(variant, "Mupit_Structure_id", mupit_structure_id)
                 variant.save()
             time.sleep(0.1)
+
+
+class Benchmark(object):
```
izcram: Nice :)

falquaddoomi:
Regarding the recent commits, they're a result of me rebasing this branch onto master today to bring it up to date; Git apparently picked up some older commits as "recent commits" due to the rebase and force-push.

Also, there were apparently some Django migrations introduced into master, so I added a merge migration that brings together the database changes in this PR (i.e., indexing various columns) with the ones in master. (I manually inspected both sets of migrations beforehand to ensure they didn't conflict.)

PR #1016, opened 2019-03-08
implement points sorting for literature search results, rem extra spa…

zfisch:
…ce in docs


zfisch:
• Sorting is now handled entirely in the server in descending order based on points.
• Papers without a year are given 0000 values, and displayed as unknown in the UI/downloads.

Issue #1029, opened 2019-03-27
Reorganize "More" Pages and Directories


amycoffin:
When looking at the "more" content, I realized that there are a lot of old files floating around. 

Two examples include: 
1. there is a "history.md" that is referenced in:
       brca-exchange/website/js/content.js
But then the same/analogous content is hardcoded into:
       brca-exchange/website/js/FactSheet.js

2. According to google analytics, there is a ghost page that gets hits, presumably through google searching for information on BRCA variation:
[Google Analytics](https://analytics.google.com/analytics/web/#/report/content-pages/a30159111w105300047p109579554/explorer-table.plotKeys=%5B%5D&_r.drilldown=analytics.pagePath:~2Fabout~2Fvariation)
[Ghost Page](https://brcaexchange.org/about/variation)

There might be more out of date content floating around, but I can't reliably trace everything through the javascript to tell what is old and what is active. 

So, I think it makes sense for a developer to do a reorganization according to [these wireframes, slides 10-17](https://docs.google.com/presentation/d/1TCtFdY2j6MTVBwBpP_TpqL36wMxVVlzo_ZEKV1b1YOg/edit#slide=id.g46a0a507e9_0_65), cleaning up and deleting any unreferenced our out of date files as they go. 

Note that in these wireframes, the "Media Inquiries" snippet has been omitted. Further, there are a few titles and headers that have changed, but otherwise all content reorganized but not changed.

amycoffin:
@zfisch is this worth doing as part of the analytics project? It would create less pages to manage overall

zfisch:
@amycoffin there's might be some crossover -- we can associate the issue regardless. removing unused files won't affect SEO but dead links definitely will.

PR #1030, opened 2019-03-27
reorganized and merged Help with FAQ

amycoffin:
Here are a list of changes that were made in this pull request:

1. copying FAQ markdown to help/default and help/research directories. There are now duplicate files in each of those directories. 
2. Changing the headers and order of help page tiles to be more logical, and to accommodate new FAQ items 
3. Changing all titles of help page tiles to be questions

Things I would appreciate a double check on:
1. Making sure that the }s, ]s, etc. stayed aligned and block together correctly. It looks fine on the locally hosted version of the site, but I'm not totally sure since I don't know js. 
2. Remove the beta tag label on "How is the data on this site updated?" (oops) 

Next steps: 
1. Delete FAQ page from UI and its underlying directory
2. Rename "Help" as "FAQ" 
3. Reorganize other pages according to issue #1029, and cleaning up directories accordingly.
4. Make changes available on sandbox for review.


website/js/content.js
```diff
@@ -272,21 +294,56 @@ const helpContentResearch = [
...
             },
+            {
+                name: "What genome build is this site using?",
+                id: "cravat-mupit-3d-protein-view",
+                contents: require("../content/help/research/what-genome-build-is-this-site-using.md")
+            },
+            {
+                name: "How is the data on this site updated?",
+                contents: require("../content/help/research/how-is-the-data-on-this-site-updated.md"),
+                isBeta: true
```
zfisch: if you remove this line `isBeta: true` it will remove the beta tag from this question

amycoffin:
@zfisch I sent some clarifying questions over slack to ensure I understand how to change this correctly through git, and then I can take care of it.

PR #1046, opened 2019-05-01
GnomAD pipeline and backend, closes #1024

### LINKED ISSUE #1024: integrate gnomad data into automated pipeline
https://gnomad.broadinstitute.org/gene/ENSG00000012048?dataset=gnomad_r2_1_non_cancer

zfisch:
Early PR that accomplishes the following:

1. Downloads data from GnomAD (their API is in flux, so we may need to use a static previously downloaded dataset if this stops working as expected).
2. Parses relevant fields, calculates required allele counts, numbers and frequencies.
3. Normalizes data.
4. Integrates into automated pipeline and variant merging process.
5. Creates fields in database and django models.

Some notes:

The GnomAD fields are the first text fields for which I've set null=True. Generally speaking, text fields should be non-nullable, but making them so requires a default value, and adding that default value produces a monster migration that crashes on our servers. If we truly want them non-nullable, we may need to add default values manually to all existing rows and then change the fields to be non-nullable afterward (see https://stackoverflow.com/questions/44167386/django-migration-11-million-rows-need-to-break-it-down). Open to thoughts on this if it's worth changing, but I'm not sure having null values is really a problem for us.

Outstanding tasks:

- [ ] General code cleanup
- [ ] Add proper gnomad annotation for vcf file
- [ ] Confirm all required fields are present (requires UI approval from @amycoffin @melissacline 
 and EGG)


pipeline/gnomad/parse_gnomad_data.py
```diff
@@ -0,0 +1,145 @@
...
+        'hgvsc',
+        'hgvsp',
+        'pos',
+        'ref',
+        'variantId',
+        'consequence'
+    ]
+
+    column_name_mapping = {
+        'pos': 'pos_hg19',
```
izcram: They don't offer hg38? Or what is the reason hg19 is still used?

pipeline/gnomad/download_gnomad_data.py
```diff
@@ -0,0 +1,154 @@
...
+    logfile = options.logfile
+
+    if options.verbose:
+        logging_level = logging.DEBUG
+    else:
+        logging_level = logging.CRITICAL
+
+    logging.basicConfig(filename=logfile, filemode="w", level=logging_level)
+
+    variants_brca1 = generate_data("BRCA1")
```
izcram: Gene name should probably be passed in as a program argument, i.e. either a complete list or just one gene. In the latter case, we would call this script for every gene separately.

pipeline/gnomad/gnomad_to_vcf.py
```diff
@@ -0,0 +1,136 @@
...
+import pyhgvs.utils as hgvs_utils
+from pygr.seqdb import SequenceFileDB
+import logging
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Convert database table to VCF format.')
+    parser.add_argument('-i', '--input', type=argparse.FileType('r'),
+                        help='Input file for conversion.')
+    parser.add_argument('-a', '--inAnnot', default='/hive/groups/cgl/brca/phase1/data/resources/gnomadAnnotation',
```
izcram: I wonder, whether it makes any sense to have default like those. I don't see any benefit except maybe for one developer working in a specific environment. It may be more useful to have paths valid in the docker image (but no default would be fine with me as well)

PR #1047, opened 2019-05-03
Feature/variant merging arbitrary genes

izcram:
Preview PR: waiting for the integration of #1046 and then rebase

Generalizes variant merging to arbitrary gene sets (as long as they cover mutually disjoint regions in the genome).

Reference sequences are now loaded from a local [sequence repository](https://github.com/biocommons/biocommons.seqrepo) instead of text files. Sequence edits of a variant are still performed wrt the entire reference sequence of a gene (`find_equivalent_variants_whole_seq`). 

An alternate strategy has been implemented where the edits happen on some smaller region ('chunk') of the genome (`find_equivalent_variant`). This strategy, however, requires to pick some safety margin to determine the potential position range on a chromosome where a variant differs from a reference sequence due to repeat regions in a genome. A wrong choice of such a margin could potentially lead to erroneous results. Since we don't intend to support overly large genes in the near future, the in principle less efficient whole sequence strategy is currently used, which doesn't require such a choice.

Also, gene boundaries are enforced now in the merging step, i.e. if a variant doesn't fall into prespecified regions, it is discarded.


pipeline/data_merging/utilities.py
```diff
@@ -1,4 +1,79 @@
...
+    as payload.
+
+    Used to return the gene symbol (e.g. BRCA1) of a given position.
+
+    :param gene_config_df:
+    :return: dict[int, IntervalTree]
+    '''
+    gene_regions_dict = extract_gene_regions_dict(gene_config_df)
+
+    interval_tree_mapper = lambda c, s, e: gene_regions_dict[ChrInterval(c, s,e)]['symbol']
```
zfisch: obv not a big deal but spacing on `[ChrInterval(c, s,e)]['symbol']` is funky

izcram:
I've addressed your comments @zfisch , thanks for looking over the code! Then I needed to fix some off-by-one bugs in order to make all variants work. The diff of `merged.tsv` between the May release data and what is generated by the code in the current PR is now empty, so we should be good to merge!

PR #1072, opened 2019-06-18
VR Data API endpoint (Work In Progress)

falquaddoomi:
This PR adds a few things:
- a model, VariantRepresentation, consisting of a Variant reference and a JSON blob called "Description", which stores the VR representation of the variant.
- an endpoint, /data/variantreps, which outputs the contents of VariantRepresentation (plus the variant's ID and Genomic_Coordinate_hg38 column).
- a management command, add_vr_rep, which takes a two-column TSV with a header ("Variant_id", "Genomic_Coordinate") as an argument and inserts it into VariantRepresentation. It joins on the Genomic_Coordinate_hg38 column, as that was what was supplied in the reference TSV.

We may want to extend the endpoint to allow querying by variant ID and/or other attributes. We may also consider paging the response if it ends up being big and our clients need it to be paged.


website/django/data/migrations/0042_variantrepresentation.py
```diff
@@ -0,0 +1,24 @@
...
+
+    dependencies = [
+        ('data', '0041_paper_crawl_date'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='VariantRepresentation',
+            fields=[
+                ('Variant', models.OneToOneField(on_delete=django.db.models.deletion.CASCADE, primary_key=True, serialize=False, to='data.Variant')),
```
zfisch: Curious if it should be a one-to-one relationship or if we just want all variants with a given genomic coordinate to point to the same VariantRepresentation. Should probably get @melissacline's input on this, but I understand the only reason to do one-to-one is to allow versioning to take VR representation into account -- is that correct @falquaddoomi ?

zfisch: From Melissa on Slack:

> Yes, the VR representation is just static data associated with the genomic coordinate.  This “static” data will evolve as the VR representation evolves.  I’m not picturing any use case for old versions of the VR representation.  So, I’d say that it doesn’t need to be versioned.

PR #1076, opened 2019-06-26
Fix/fix boundaries in variant merging

izcram:
So far, no variants were filtered during the variant merging. But after the latest tweaks to the merging code, this step does not work with out-of-bounds variants hence they need to be filtered out. However, this filtering was not working properly which made the pipeline crash with more restrictive boundaries. This should be fixed now.


izcram:
@zfisch with the latest fixes we get the expected numbers again:

(variants by source wrt to May release)
```
{'ClinVar': 17136, 'Findlay_BRCA1_Ring_Function_Scores': 3893, 'ENIGMA': 7256, 'ESP': 688, '1000_Genomes': 4349, 'ExAC': 3362, 'exLOVD': 295, 'BIC': 3618, 'LOVD': 8493}
{'changed_information': 3770, 'new': 582, 'added_information': 493, 'changed_classification': 425}
```

PR #1081, opened 2019-07-23
Gnomad

zfisch:
1. Downloads and parses gnomAD data from gnomAD api (might need to be made static, but we can see how long this is operable). Runs through automated pipeline and merges into output.

2. Adds fields to DB and serves through API along with all variant/report requests.

3. Displays graphs and numerical data in allele frequencies tile.

Todo:

- [x] Allele frequency numbers differ from gnomAD website. All we're doing is pulling the data through gnomAD's api and rounding to 4 sig figs, but we should figure out if this is an issue and where the issue lies.

- [x] It seems empty graphs tiles are still displayed in the allele frequency tile. Should hide by default.

- [x] Needs review before merge.

- [x] More tests could be written.

- [x] Final code cleanup.


zfisch:
@falquaddoomi @izcram this PR is ready for review if either of you would like to take a look -- planning to merge and deploy once the data is approved

PR #1101, opened 2019-08-26
fixing segfault with apache after upgrading libssl to 1.1.1 on host

izcram:


falquaddoomi:
FYI, to install this on my OS X machine I had to specify the location of libssl like so:

```bash
# after having activated the virtualenv, pwd: brca-exchange/website
pip install --global-option=build_ext \
            --global-option="-I/usr/local/opt/openssl/include" \
            --global-option="-L/usr/local/opt/openssl/lib" -r requirements.txt
```

(Found the above on https://stackoverflow.com/a/41801121/346905)

PR #1108, opened 2019-09-03
Upgrade web backend to python3

zfisch:
This PR adds python3 support and drops python2 support for the web backend. It would be super helpful to get some extra eyes on this, @falquaddoomi especially to make sure it doesn't cause any issues with the mobile app.

Note: mysqlclient was causing issues for python3, I vaguely remember @falquaddoomi adding it as a dependency a while back but don't remember the purpose. Is it safe to remove?


falquaddoomi:
Hey, yes, I believe it's safe to remove mysqlclient. It was added by Chloe, actually, in 502701afb182139f10d131281435612d87c05422, but the splicing pipeline later ended up getting dockerized. The mysqlclient dependency is still present in https://github.com/BRCAChallenge/brca-exchange/blob/6cbe801721/pipeline/splicing/requirements.txt and it's apparently used in the splicing pipeline to fetch genomic coordinates from a UCSC-managed MySQL server here: https://github.com/BRCAChallenge/brca-exchange/blob/6cbe801721/pipeline/splicing/calcMaxEntScanMeanStd.py#L30

website/js/NavBarNew.js
```diff
@@ -119,7 +119,7 @@ var NavBarNew = React.createClass({
         );
 
         return (
-            <div className={classNames("navbar-container", {"beta": this.state.isBeta})}>
+            <div className={classNames("navbar-container", {"beta": this.state.isBeta && false})}>
```
falquaddoomi: Just checking, but since this expression will always be false, was it the intent to disable the beta switch?

website/django/data/views.py
```diff
@@ -274,7 +275,7 @@ def index(request):
     if release:
         query = Variant.objects.filter(Data_Release_id=int(release))
         if(change_types):
-            change_types = map(lambda c: change_types_map[c], filter(lambda c: c in change_types_map, change_types))
+            change_types = [change_types_map[c] for c in [c for c in change_types if c in change_types_map]]
```
falquaddoomi: I assume `2to3` generated this code, and if so it's pretty impressive that it can convert a map and a filter into a list comprehension with an if clause. Kudos to them.

website/django/data/management/commands/add_papers.py
```diff
@@ -27,7 +27,7 @@ def handle(self, *args, **options):
         existing_variant_papers.update(deleted=True)
 
         paper_objects = {}
-        for pmid, paper in papers.iteritems():
+        for pmid, paper in list(papers.items()):
```
falquaddoomi: I assume these changes where a generator is wrapped in a call to `list()` are introduced by `2to3`. It's not wrong per se, but it's a surprisingly conservative choice on their part to convert every generator to a list -- in the cases I've seen later in the file they'd all be fine since they're being iterated over anyway. Anyway, no action needed, but I thought it was strange.

website/js/NavBarNew.js
zfisch: ah good catch, this was a temporary change for testing. will fix

website/django/data/views.py
zfisch: yep.. pretty impressive

Issue #1090, opened 2019-08-06
Ctrl + F functionality doesn't work on Help pages


amycoffin:
When we implemented tile functionality on the home page, it made it so that Ctrl +F no longer works. We should implement a way for people to search the help pages.

amycoffin:
One starting point is perhaps auto-expanding anything with sub-tiles, so at least titles of subtiles are able to be found using ctrl + F

falquaddoomi:
Closed by PR #1112.

PR #1112, opened 2019-09-17
Adds search box, hardlink helpers to help page

falquaddoomi:
Adds a search box with navigation functionality to the Help page, since the regular browser search can't see the contents of collapsed panels (as documented in #1090).

Also adds "hardlink helpers", initially-hidden buttons that appear when you hover over a linkable item, e.g., a header or definition. Clicking the button changes the address bar to the URL of the linkable item and copies the URL to the user's clipboard. Note that these intentionally do not cause a navigation event, since that would cause the page to scroll. The RawHTML component now includes an extra boolean parameter, "hardlinks"; if enabled, it searches for linkable elements in its body text and appends hardlink helpers for each.


website/js/Help.js
```diff
@@ -2,22 +2,32 @@
 'use strict';
 
 import classNames from "classnames";
+import update from 'immutability-helper';
 import BetaTag from "./components/BetaTag";
+import {debounce} from "lodash";
+import SearchController, {EXTRA_SEARCH_PADDING} from "./components/help/SearchController";
+import HardlinkHelper, {idHelpClicked} from "./components/help/HardlinkHelper";
+// import debounce from 'lodash/debounce';
```
zfisch: can remove?

PR #1120, opened 2019-10-01
API Docs + Notebook

falquaddoomi:
This PR adds some documentation for the internal BRCA Exchange API. Currently, the docs consist of a markdown file with details about the endpoints, and a notebook that contains a tutorial for querying the API from Python.


falquaddoomi:
I'm still not sure where this documentation should live; I arbitrarily chose `/website/djano/data/static` beacuse that's where the previous notebook was located, but I'm all ears for other locations to put it.

PR #1127, opened 2019-10-15
Updates API links, moves docs out of repo

falquaddoomi:
This PR updates the links on the API page to the new docs I wrote a little while ago. It also moves the docs out to a separate repo, described below.

[Binder](https://mybinder.org/), the interactive notebook hosting service I'm using requires that the notebook's requirements.txt file live in the root of the repo, which is incompatible with our main repo's structure. Due to that, I decided to (temporarily?) move the tutorial notebooks and API docs to https://github.com/BRCAChallenge/brca-exchange-api-docs.


zfisch:
Decided to forgo Binder and keep the docs in the repo for simplicity.

PR #1150, opened 2019-11-22
Hgvs nomenclature

zfisch:
Determines and adds new hgvs nomenclature to the database. If the new nomenclature is available, it's displayed instead of the old Genomic Coordinate values. If not, the old values are displayed.


pipeline/data_merging/brca_pseudonym_generator.py
```diff
@@ -85,6 +83,16 @@ def from_field_or_compute(row):
     return s_cdna
 
 
+def compute_genomic_hgvs(cDNA, assemblyMapper):
+    try:
+        genomic_hgvs = assemblyMapper.c_to_g(cDNA)
+        return str(genomic_hgvs)
+    except hgvs.exceptions.HGVSInvalidIntervalError as e:
+        logging.info("HGVSInvalidIntervalError: The given coordinate is outside the bounds of the reference sequence: " +
+                     str(cDNA))
```
izcram: I'd catch the `HGVSError`, which is a base class for all sorts of HGVS issues. Also, I'd give some generic message along with the `cDNA` with the message from the actual error, like so:
```
logging.info("Exception during conversion of " + str(cDNA) + " to genomic coordinates: " + str(e)
```

website/js/DataTable.js
```diff
@@ -37,22 +38,31 @@ function setPages({data, count, deletedCount, synonyms, releaseName}, pageLength
...
+        return variantData;
+    },
+    determineGenomicCoordinates: function(variantData) {
+        let field = "Genomic_Coordinate_hg37";
+        let hgvs = variantData.Genomic_HGVS_37;
+        variantData = this.normalizeGenomicCoordinate(field, hgvs, variantData);
+
+        field = "Genomic_Coordinate_hg38";
+        hgvs = variantData.Genomic_HGVS_38;
+        variantData = this.normalizeGenomicCoordinate(field, hgvs, variantData);
```
izcram: not sure what the intention is, but it seems the `Genomic_Coordinate_hg37` related variables are just overwritten

zfisch: exactly -- the purpose of this PR is to introduce a new standardized variant nomenclature and use it if it's available. if not, we use the old Genomic_Coordinate_hg38/Genomic_Coordinate_hg37 values.

pipeline/data_merging/brca_pseudonym_generator.py
zfisch: good call, this is now resolved :)

PR #1175, opened 2020-02-04
Upgrade frontend libs step 1: react-bootstrap

falquaddoomi:
This PR primarily makes changes to allow us to upgrade react-bootstrap to the most recent non-BS4 version (0.33.1). It also updates other dependent libraries for compatibility with React 0.14. I intend to upgrade to the most recent version of React (v16), although that will require some careful editing and testing.


zfisch:
@falquaddoomi some more issues:

- [ ] Switching from detail to summary view and back causes the site to freeze
- [ ] Clicking "Show detail view" on the /variants page no longer works

I also recommend opening the console as you navigate around the site. There are some new errors and warnings that aren't present on production.

PR #1182, opened 2020-02-21
ignore None type synonym values in clinvar

zfisch:
@izcram ClinVar parsing was ending prematurely because a None value was being inserted into the list of synonyms. I'm assuming if a.text is None it's safe to ignore that value as a synonym and move on, is this correct?

If not, could you dig in a little deeper and help us get this sorted? It's blocking pipeline runs at the moment.

Thanks!


izcram:
Yes, looks like empty elements like the following were recently added to clinvar:
```
<Attribute Change="p.Arg332=" Type="HGVS, protein"/>
```

As they contain no accession we better ignore these elements (we get the information we want from other Attribute elements anyways)

PR #1184, opened 2020-02-25
Feature/replace pyhgvs and pygr in to vcf scripts

izcram:
With this PR, conversions to genomic coordinates during the staging of VCF files are performed using the biocommons `hgvs` library replacing `pyhgvs`. In the case of findlay assays and gnomad, the genomic coordinates are now taken from the source instead of converting the cDNA coordinates.

Here a summary of changes in data by affected source:
- findlay assays: no changes
- gnomad: changes in genomic coordinates for ~20 variants
- exLOVD and LOVD: additional duplication (`dup`) ~170 variants


pipeline/common/vcf_files_helper.py
```diff
@@ -0,0 +1,40 @@
...
+import hgvs
+from seq_utils import SeqRepoWrapper
+
+
+def cdna_str_to_genomic_var(cdna_hgvs_str, assembly=HgvsWrapper.GRCh38_Assem, hgvs_wrapper = HgvsWrapper.get_instance(),
+                            seq_fetcher=SeqRepoWrapper.get_instance()):
+    """
+    Convert a CDNA HGVS string into a genomic coordinate.
+    Normalizes in genomic space.
+    Note, that the seq_fetcher needs to be in the same assembly as given by the `assembly` argument.
```
zfisch: is there a way to enforce a check on this? e.g. assert statement

izcram: yes, good point! Fixed in 1a5201dc7524c5f7459048dd14d8b9d24eea32d4

zfisch:
Output looks great! Merging :)

PR #1186, opened 2020-02-27
Vr api

### LINKED ISSUE #1166: integrate VR data into dB and provide api to access variants by VRID

zfisch:
Closes #1166 

1. Adds a dockerized script to gather and append VR_ID's to variants (when available).
2. Includes the vr_id script as part of the automated pipeline.
3. Adds VR data to release and database.
4. Added endpoint to retrieve variant data by vr_id.


pipeline/vr/appendvrids.sh
```diff
@@ -0,0 +1,38 @@
...
+  --name append-vr-ids \
+  --network=host \
+  -v "${ARTIFACT_DIR}":/artifacts \
+  -e UTA_DB_URL \
+  "${IMAGE_NAME}" \
+  -i "/artifacts/${INPUT_FILE}" \
+  -o "/artifacts/${OUTPUT_FILE}"
+
+# --- post-step: when vr's done, bring down its dependent services
+docker stop seqrepo-service
```
izcram: This can lead to unexpected results if there are 2 pipelines running on the same machine. I would just leave the service running

pipeline/vr/appendvrids.sh
```diff
@@ -0,0 +1,38 @@
...
+IMAGE_NAME="${4:-brcachallenge/append-vr-ids:0.1}"
+SEQ_REPO_DIR="${5:-/usr/local/share/seqrepo}"
+
+# --- pre-step: build vr image (caching will make this fast if it was already built)
+docker build -t "${IMAGE_NAME}" .
+
+# ----------------------------------------------------
+# --- 1. spin up seqrepo-rest-service
+# ----------------------------------------------------
+docker run --rm -d \
```
izcram: this should probably be done via the `start-seqrepo-rest-service`

izcram:
pipeline run on a different machine went through!

pipeline/vr/README.md
```diff
@@ -0,0 +1,8 @@
+appendVRIds was dockerized in order to run using py3 (https://github.com/ga4gh/vr-python requires python 3) while the existing pipeline is still running on py2
```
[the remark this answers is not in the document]
zfisch: definitely

pipeline/vr/appendvrids.sh
falquaddoomi: Agreed. For that matter, starting the service will fail if there's already an instance of it running, since it starts in host networking mode and would block the second instance from listening on port 5000. As you mentioned elsewhere, I suppose the service should be started just once via the `start-seqrepo-rest-service` target. If it fails to start due to the port already being bound (e.g., if a second pipeline run was initiated), then I suppose that failure should be ignored.

zfisch: ya.. im going to test this now. I'm planning to keep the docker run command to start the seqrepo-rest-service in appendvrids.sh so that it could hypothetically be run in isolation. I'm assuming that a failure to start the service because one is already running won't cause any issues.

pipeline/vr/appendvrids.sh
zfisch: testing keeping this here so it can run in isolation...
if it fails because one's already running, that seems harmless

pipeline/Makefile
```diff
@@ -38,6 +40,9 @@ download-resources: ## download resources files
 start-local-uta: ## starting local uta docker container
 	[ `docker ps -f name=$(UTA_CONTAINER) | wc -l` -gt 1 ] || docker run -dit --name $(UTA_CONTAINER) -p $(UTA_PORT):5432 $(UTA_DOCKER_IMAGE)
 
+start-seqrepo-rest-service:
```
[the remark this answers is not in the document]
zfisch: done

zfisch:
@izcram @falquaddoomi requested changes have been made, except for keeping the appendVRIds as non-pandas
let me know if there's anything else you'd like to see changed before merge

PR #1188, opened 2020-03-10
Feature/remove obsolete code

izcram:
Also removing pyhgvs and pygr as a dependency.

This PR removes some code which doesn't seemed to be used by the pipeline nor anybody else. The code would still be available in the repository history, however, not visible anymore.


zfisch:
Looks good! Merge whenever you're ready :)

PR #1199, opened 2020-04-20
Update to include the gnomAD subpopulations

melissacline:
I'm wondering if some of these lines are getting too long...


zfisch:
@melissacline 

We also added these fields:

exome_popmax
exome_popmax_population
genome_popmax
genome_popmax_population
hgvs
variantId

and dropped these fields:

variant_id
hgvsp
hgvsc

I believe the new hgvs field is the same as the hgvsc field, and variantId is the same as variant_id.

Could you provide brief descriptions for the popmax fields?

melissacline:
@zfisch Thanks!  I'll work on the descriptions for those additional new fields.  Do they all have a GnomAD suffix, e.g. hgvs_GnomAD?

zfisch:
no suffixes in this file!

zfisch:
as long as we have descriptions for all the new fields i can clean up the rest 👍

PR #1215, opened 2020-05-15
Feature/verify exit code

izcram:
This PR hopefully make pipeline operations a bit more robust by
1) explicitly checking the return/exit code of a subprocess to stop the pipeline if it reports that it failed
2) prefixing output files of a failed subprocess with FAILED* , s.t. they can be inspected, but implicitly block the pipeline from running further


zfisch:
The above error should now be resolved

izcram:
no, I'd need a week more. I try to give this a priority next week, s.t. we can wrap up all these dangling pieces by the end of the month at the latest

PR #1218, opened 2020-05-29
Adding the VR_ID and Genomic_HGVS_38 and 38 fields.  Reordering for c…

melissacline:
…onsistency with built_with_change_types.tsv/


pipeline/built_with_change_types.dictionary.tsv
```diff
@@ -1,5 +1,5 @@
 Field Name	Description	Example Value	Useful Resources
-Source	One or more labs, repositories, consortiums, or other entities who reported data for the variant (does not include Submitters to ClinVar)	ENIGMA,ClinVar,LOVD,ESP,1000_Genomes,ExAC,BIC,exLOVD
```
zfisch, line 2: was BIC removed here intentionally? how come?

melissacline: Yes, I removed it intentionally because it's deprecated as a data source.

zfisch: I'm unclear... we still collect it in our pipeline runs, label variants as such in the database, and provide a filter in the UI for it

melissacline: Hmm, I see what you mean.  For consistency, we should either include BIC here or exclude it everywhere.  I'll add it back in.

melissacline: Duly updated.

PR #1227, opened 2020-07-09
fixes pipeline failures due to None cDNA values -- also logs rows wit…

zfisch:
…hout calculable cDNA values

This PR prevents pipeline failure / exit when cdna values are not available. It also logs rows for which cDNA values cannot be calculated.


pipeline/data_merging/brca_pseudonym_generator.py
```diff
@@ -75,7 +75,10 @@ def from_field_or_compute(row):
         computed = compute_hgvs(row)
 
         if not computed:
-            return cdna_from_cdna_field(row)
+            cDNA = cdna_from_cdna_field(row)
+            if cDNA is None:
+                logging.info("No calculable cDNA for row: " + str(row))
```
izcram, line 80: since we moved to python3, you can also use string interpolation like this `f"No calculable cDNA for row: {row}"`

pipeline/data_merging/brca_pseudonym_generator.py
```diff
@@ -100,6 +103,9 @@ def compute_genomic_hgvs(cDNA, assemblyMapper):
     except HGVSError as e:
         logging.info("Exception during conversion of " + str(cDNA) + " to genomic coordinates: " + str(e))
         return None
+    except AttributeError as e:
```
izcram, line 106: I think it would be a cleaner fix to just test if cdna is None at the beginning of the function. This makes it clear that we expect that `cDNA` can be `None` and how we handle this case. (you get this AttributeError whenever `cDNA` is `None`)

pipeline/data_merging/brca_pseudonym_generator.py
```diff
@@ -77,7 +77,7 @@ def from_field_or_compute(row):
         if not computed:
             cDNA = cdna_from_cdna_field(row)
             if cDNA is None:
-                logging.info("No calculable cDNA for row: " + str(row))
+                logging.info("No calculable cDNA for row: {row}")
```
izcram, line 80: need the f in front in order for the string interpolation to work: `logging.info(f"No calculable cDNA for row: {row}")`

PR #1250, opened 2020-08-27
Luigi config files

melissacline:
Added a configuration file for the full suite of genes in the BRCA and HBOP purview


izcram:
Thanks for providing the config files!

2 things:

- the `luigi` folder got renamed to `workflow` recently (avoiding confusion between "our" luigi python module and the luigi package -> better support in IDEs), could you move the files accordingly?
- while you are at it, could you add a column `strand` at the end and fill in `+` and `-` accordingly? (won't be used as of now, but then once the strand handling code is ready this is already done)

melissacline (commented):
Hi @izcram, the content looks fine in terms of accuracy, but in format I'm a bit bemused why you chose to use the words "plus" and "minus" rather than symbols.  One imagines that this was the results of a well-reasoned decision, but what exactly was your reasoning?  Thanks!

izcram:
@melissacline unfortunately we already use the '-' sign to mean that a value is missing (na value) in the config files (and also more generally in the files produced by the pipeline). We could change this and make the parsing of the config file a bit more involved, but for the sake of consistency I opted for the non-symbol approach.

melissacline:
Gotcha!  That makes perfect sense.  All the changes look good.

PR #1290, opened 2021-02-10
Multi gene pipeline

zfisch:
This PR allows user to specify a gene config file in the generate_release command. The config file determines which genes to collect data for. Downstream changes were made to support the download, aggregation, and normalization of all genes specified in the gene config file.

Existing config files were also updated to represent correct data for included genes.


pipeline/data_merging/brca_pseudonym_generator.py
```diff
@@ -231,16 +238,24 @@ def main(input, output, log_path, config_file, resources, processes):
...
     logging.info("Compute hg37 normalized representation of internal")
     # normalizing again for the hg37 representation. An alternative would be to convert the normalized hg38 representation to hg37.
     # If we use crossmap, we would need a way to convert the VCF like representation back to an hgvs object, which we currently
     # are unable to do properly. That is, we can use VCFVariant.to_hgvs_obj, however, structural variants will be converted
     # to delins, losing information if a variant was e.g. a del, ins, or dup.
 
-    df[TMP_HGVS_HG37] = pd.Series(
-        [v.to_hgvs_obj(hgvs_proc.contig_maps[HgvsWrapper.GRCh37_Assem]) for v in var_objs_hg37])
+    # set None values for any hg37 coordinates that cannot be derived and add to lists in proper order
+    tmp_hgvs_hg37_values = [v.to_hgvs_obj(hgvs_proc.contig_maps[HgvsWrapper.GRCh37_Assem]) for v in var_objs_hg37]
```
izcram, line 250: would be good to move this block _before_ the `logging.info("Compute hg37 normalized representation of internal")` line, as it's not about normalization but merely about computing a hg37 representation.
It would be also good to put this into a separate function for better readability of the main function.

pipeline/data_merging/brca_pseudonym_generator.py
```diff
@@ -231,16 +238,24 @@ def main(input, output, log_path, config_file, resources, processes):
...
+    for v in var_objs_hg37_failed:
+        hgvs_obj = v.to_hgvs_obj(hgvs_proc.contig_maps[HgvsWrapper.GRCh38_Assem])
+        row_number = df[df[GENOMIC_HGVS_HG38_COL] == str(hgvs_obj)].index[0]
```
izcram, line 253: instead recomputing the hgvs_obj you could compare to the `VAR_OBJ_FIELD`:
```
row_number = df[df[VAR_OBJ_FIELD] == v].index[0]
```
This is simpler and probably faster.

pipeline/workflow/gene_config_all_genes.txt
```diff
@@ -0,0 +1,11 @@
+symbol,entrez_id,chr,end_hg38,ensembl_id,start_hg38,start_hg38_legacy_variants,end_hg38_legacy_variants,synonyms_ac_col,hgvs_cdna_default_ac,start_hg37,end_hg37,strand
```
izcram, line 1: I ran a
```
 cat gene_config_*.txt  | grep -v -e '^symbol,' | sort | uniq
```
to see if for every gene there is only one line, that is, the information is consistent across all genes.
Turns out, that  `CHEK2` and `PALB2` in `gene_config_extended.txt` are different from e.g. in this file. I wonder which one it is we want. Also perhaps the `gene_config_extended.txt` may be obsolete now, so we could remove it.

pipeline/workflow/pipeline_utils.py
```diff
@@ -117,6 +119,17 @@ def concatenate_symbols(symbols):
     return concatted_symbols
 
 
+def concatenate_files_with_identical_headers(file_directory, output):
```
izcram, line 122: A minor thing, but in the spirit to keep the code base concise, one could also do (unless there is a subtlety I miss):
```
 subprocess.run(f'cat {file_directory}/*.txt > out', shell=True)
```

izcram (changes requested):
Very nice! :)
I made a few suggestions which mostly are about making the code more readable/compact

pipeline/data_merging/brca_pseudonym_generator.py
zfisch: This was my first approach too :) In theory it would work, but there's an issue with formatting (notice the '' around the chr value):
```
(Pdb) df[VAR_OBJ_FIELD].iloc[9]
VCFVariant(chr=11, pos=108312440, ref='G', alt='A')
(Pdb) v
VCFVariant(chr='11', pos=108312440, ref='G', alt='A')
(Pdb) df[df[VAR_OBJ_FIELD] == v].index[0]
*** IndexError: index 0 is out of bounds for axis 0 with size 0
```

For what it's worth, there's only one problematic variant of >80,000 at the moment. I think we can take the performance hit of recomputing hgvs_obj 😁

pipeline/workflow/pipeline_utils.py
zfisch: yea... this function throws out the first line (header row) of every file after the first

pipeline/data_merging/brca_pseudonym_generator.py
```diff
@@ -284,7 +299,17 @@ def _compute_cdna(df_part):
     df[PYHGVS_GENOMIC_COORDINATE_38_COL] = df[VAR_OBJ_FIELD].apply(lambda v: str(v))
 
     df[PYHGVS_GENOMIC_COORDINATE_37_COL] = pd.Series([str(v) for v in var_objs_hg37])
-    df[PYHGVS_HG37_START_COL] = pd.Series([v.pos for v in var_objs_hg37])
+
+    # handles missing hg37 coordinates
+    hg37_start_col_pos = []
+    for v in var_objs_hg37:
```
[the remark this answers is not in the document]
zfisch: I was confused why i couldn't get that to work before, the correct syntax is:

`hg37_start_col_pos = [ v.pos if v else None for v in var_objs_hg37 ]`

pipeline/Makefile
```diff
@@ -119,6 +124,7 @@ run-task: ## Running a specific task
 PIPELINE_ARGS := --PipelineParams-output-dir /files/data/output --PipelineParams-file-parent-dir /files/data --PipelineParams-date $(DATA_DATE) --PipelineParams-gene-config-path $(GENE_CONFIG_FILE)
 
 force-run-task: ## Running a specific task (deleting its dependencies first)
+	## TODO: should this command use RUN_ARGS command twice?
```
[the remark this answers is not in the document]
zfisch: interestingly, removing one changes the output and throws a new error. i'm going to leave as is with this todo since it's irrelevant to the changes in this PR

pipeline/workflow/gene_config_all_genes.txt
zfisch: updated! i left all config files in place after updating content in case they prove useful later, can also remove if we don't need them

zfisch:
@izcram I went through and made updates as requested or left comments with explanations why I chose to leave them as is. Let me know if there's anything else :)

pipeline/data_merging/brca_pseudonym_generator.py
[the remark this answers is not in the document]
izcram: yes of course! sorry for leading you on a wrong track

izcram (approved):

PR #1291, opened 2021-03-05
Feature/add variants output file

izcram:
Adds in new task GenerateVariantsOutputFile to generated a simplified version of the full pipeline output (output/release/built_final.tsv) with the goal to be more easily consumable for an analyst or a downstream application.

Includes a jupyter notebook with some example usage.


zfisch (approved):
awesome work! looks good to me 👍

Issue #1292, opened 2021-03-09
clean up tarball


zfisch:
which files do we want to include? all others can be excluded from the tarball

zfisch:
focus on cleaning up the artifacts directory -- can throw out everything except logfiles, discarded_reports.tsv, and wrong_genome_coors files

PR #1300, opened 2021-04-06
Feature/cleanup tar ball

izcram:
addressing #1292 by introducing a `tarball_files_keep_list.txt` file enumerating which files should be added in the final tarball. Additionally also add a `tarball_files_discard_list.txt` file to be able to warn when an unaccounted file in the release dir is found (which could mean, that it was added in a pipeline iteration, but was forgotten to be added to the tarball)

Furthermore, changed output of the md5sum generation step to be compatible with GNU md5sum. This way, file integrity can be checked using `md5sum -c md5sums.txt`


melissacline:
The files to discard list looks great, and that's going to make such a difference!  
In files to keep, there are a couple that maybe we don't need to keep.

- In the artifacts directory, e have both merged.tsv and aggregated.tsv.  Would you remind me of the difference between those?  I suspect we only need one of the two.
- Would you also remind me of the difference between built_with_change_types.tsv and built_final.tsv?  I apologize that I should remember this, since we discussed it only a couple weeks ago...  built_final.tsv has no change type field.  Is there any other difference?  If not, these seem redundant, and we should pick one of the two.

Thanks!

izcram:
@melissacline 
1. yes indeed! fixed in c7088e6bef51289140667c0c7526c54d58b5c0c6
2. in 05e8f92 I changed the symlink of built_final.tsv to point to built_with_change_types.tsv mainly to make it clear that this is the final output. So it is redundant, but doesn't use more space :) I adjusted the top_level_readme.txt accordingly

zfisch (approved):
this is great!

PR #1302, opened 2021-04-16
Gnomad three point one

zfisch:
1. Gathers gnomAD 3.1 data -- now stored in a static file
2. Adds gnomAD 3.1 data to database and UI
3. Removes display of gnomAD 2.1 genome data, retains exome data
4. Updates to gnomAD data display and documentation
5. Includes an update to gather G1K data from a static file instead of regenerating in each pipeline run


pipeline/workflow/pipeline_common.py
```diff
@@ -138,12 +138,26 @@ def requires(self):
...
     def run(self):
-        pipeline_utils.create_path_if_nonexistent(
-            os.path.dirname(self.output().path))
+        if isinstance(self.output(), dict):
+            for k, v in self.output().items():
+                pipeline_utils.create_path_if_nonexistent(
+                    os.path.dirname(v.path))
+
+                copy(self.req_task.output()[k].path, v.path)
+                pipeline_utils.check_file_for_contents(v.path)
```
izcram, line 157: It seems these line could be extracted into a function and then you wouldn't have almost duplicate code in `else` branch

pipeline/workflow/gnomad_processing.py
```diff
@@ -19,85 +19,75 @@
...
-        os.chdir(gnomAD_method_dir)
-
-        args = ["python", "download_gnomad_data.py", "-o", self.output().path, "-l", self.artifacts_dir + "/gnomAD_download.log"]
-
-        pipeline_utils.run_process(args)
-        pipeline_utils.check_file_for_contents(self.output().path)
-"""
+        data = pipeline_utils.urlopen_with_retry(self.gnomAD_v3_static_data_url).read()
+        with open(self.output()["v3"].path, "wb") as f:
+            f.write(data)
```
izcram, line 41: this could probably also be extracted into as separate function to avoid some redundancy

pipeline/workflow/gnomad_processing.py
```diff
@@ -1,6 +1,6 @@
 import logging
 import os
-
+import tarfile
```
izcram, line 3: I don't think it's needed

pipeline/gnomad/download_gnomad_threepointone.py
```diff
@@ -0,0 +1,406 @@
...
+    # organize and combine variants
+    brca1_variants_non_cancer = getVariants(brca1_transcript, brca1_gene, non_cancer_dataset, reference)
+    variants_dict = convertSetToDict(brca1_variants_non_cancer)
+    brca2_variants_non_cancer = getVariants(brca2_transcript, brca2_gene, non_cancer_dataset, reference)
+    variants_dict.update(convertSetToDict(brca2_variants_non_cancer))
+
+    # find hgvs, flatten, convert to dataframe, compute allele frequencies, and normalize
+    variants_with_hgvs = find_correct_hgvs(variants_dict, (brca1_transcript, brca2_transcript))
+    variants_with_flattened_populations = flatten_populations(variants_with_hgvs)
+    variants_df = pd.json_normalize(variants_with_flattened_populations)
```
izcram, line 373: good usage of pandas API :)

pipeline/gnomad/download_gnomad_threepointone.py
```diff
@@ -0,0 +1,406 @@
...
+    exonic_variants_non_cancer = transcript_to_variants(transcript, dataset, reference)
+    for variant in exonic_variants_non_cancer:
+        variant_id = variant['variant_id']
+        if variant_id not in variant_ids:
+            variant = fetch_data_for_one_variant(variant_id, dataset)
+            exonic_set.add(json.dumps(variant, sort_keys=True))
+            variant_ids.append(variant_id)
+    retries = 0
+    max_retries = 5
+    while retries < max_retries:
```
izcram, line 232: I think it would have been nice to separate the retry logic from the querying code and the code processing the result of the query. It would be easier to follow the code

pipeline/gnomad/download_gnomad_threepointone.py
```diff
@@ -0,0 +1,406 @@
...
+            if response.ok:
+                time.sleep(0.1)
+                parse = json.loads(response.text)
+                print(f'success for variant {variant_id}')
+                return(parse['data']['variant'])
+            else:
+                print('rate limit reached, waiting 60 seconds...')
+                time.sleep(60)
+                continue
+        except json.decoder.JSONDecodeError:
```
izcram, line 189: I wonder whether it is worth to retry in all exceptions cases. Also, you may want to take a closer look at the http error codes the API give. E.g. it may only be worth to retry in case of HTTP 429 ("Too Many Requests")

pipeline/gnomad/download_gnomad_threepointone.py
```diff
@@ -0,0 +1,406 @@
...
+        try:
+            # https://stackoverflow.com/questions/49064398/requests-exceptions-chunkedencodingerror-connection-broken-incompleteread0
+            response = requests.post(
+                'http://gnomad.broadinstitute.org/api',
+                json={ "query": v_q,
+                       "variables": {}},
+                headers={"content-type": "application/json"})
+            if response.ok:
+                time.sleep(0.1)
+                parse = json.loads(response.text)
```
izcram, line 182: perhaps worth saving the request json and response json to a file which could serve as a cache. This way, if for some reason you need to restart your script, it doesn't have to start from scratch

izcram (changes requested):
some nitpicks and some more fundamental thoughts for perhaps a subsequent iteration.

zfisch:
@izcram thanks for the feedback! i did some consolidation to remove duplicate code per your suggestions.

I left the code as is inside download_gnomad_threepointone.py because that file is there for historical purposes only -- it doesn't actually get run now that we have static data. Eventually it might be good to remove it altogether or use it dynamically.

PR #1363, opened 2021-10-05
Features/variant scoring

izcram:
Prototype implementation of variant scoring algorithm.
It is a small pipeline separate from the brca exchange data pipeline, although some parts of the brca exchange code base is reused.
See more in [README.md](pipeline/gnomad/variant_scoring/README.md)


zfisch (commented):
Awesome! I can't speak much to the logic behind the scoring but the code structure is great :) Just a couple comments.

PR #1487, opened 2024-01-19
feat: URLs now default to stable CA_IDs instead of latest variant ID

e-t-k:
**Previously**: 
variant URLs are variant/12345
an older-version URL would redirect to variant/12345?redirectedFrom=123 
ClinGen Allele Registry IDs (CA_IDs)  could be used in the URL, eg variant/CA98765 , but would redirect to variant/12345.

**With this change**:
variant/CA98765 does not redirect.
variant/12345 (latest id) redirects to variant/CA98765 
variant/123 (older id) redirects to variant/CA98765?redirectedFrom=123

If a latest id doesn't have a CA_ID (eg, the variant was deleted), fallback to the original behavior instead.

The noRedirect=true param can still be added to prevent any of these redirects.

### DIFF
--- a/website/js/index.js
+++ b/website/js/index.js
@@ -621,17 +621,27 @@ var VariantDetail = React.createClass({
             });
         }
 
-        // ensure that we're viewing the latest version of the variant, and redirect if we're not,
+        // ensure that we're viewing the latest version of the variant, with the stable CA_ID URL.
+        // and redirect if we're not,
         // unless the querystring param 'noRedirect=true' is specified, in which case we stay here.
+        // In the rare case there is no CA_ID, redirect to the latest version by numeric ID instead.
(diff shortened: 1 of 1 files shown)


e-t-k:
Currently in the database there are 1976 variants without a CA_ID ; all of them are deleted. 

```
 select count(id) from currentvariant where "CA_ID" is NULL ;
 count 
-------
  1976

select distinct "CA_ID", "Change_Type_id" from currentvariant where "CA_ID" is NULL ;
 CA_ID | Change_Type_id 
-------+----------------
       |              2

select * from data_changetype;
...
  2 | deleted
```

melissacline:
This software change looks straightforward, and this functionality will be great!

PR #1522, opened 2024-08-30
Update RELEASES.md

melissacline:
Added more detail on the deployment and data deployment processes


e-t-k (approved):
I made a couple of edits but otherwise looks great!

PR #1648, opened 2026-08-20
GL community rework

GeoffLyle:
Removes data we no longer want to collect/display:

- Email Address (display only)
- User uploaded image
- Phone number

Removes need for email authentication when signing up to join the BRCA Exchange Community.

Adds fixes to the django calls so that this data is no longer delivered as part of the user information request.

TODO: Test the SignIn once actual MailChimp is implemented (needed to send out information to community members).  Test will also include whether the sign up for email updates is implemented properly.
TODO: Develop community users schema for database.

### DIFF
--- a/django/users/views.py
+++ b/django/users/views.py
@@ -316,20 +316,13 @@ def users(request):
                         )
     query = query.filter(search_query)
 
-    whitelist = ['id','email','firstName','lastName','title','role','role_other','institution','city','state','country','phone_number','hide_number','hide_email','is_active','is_admin','has_image','is_approved']
+    whitelist = ['id','firstName','lastName','title','role','role_other','institution','city','state','country','is_active','is_admin','is_approved']
 
     count = query.count()
 
     start = page_num * page_size
     end = start + page_size
     data = list(query[start:end].values(*whitelist))
-
-
-    for user in data:
-        if user['hide_email']:
-            user['email'] = ""
-        if user['hide_number']:
-            user['phone_number'] = ""
 
     response = JsonResponse({'data': data, 'count': count})
     response['Access-Control-Allow-Origin'] = '*'
@@ -350,7 +343,7 @@ def user_locations(request):
                         )
     query = query.filter(search_query)
 
-    fields = ['id', 'firstName', 'lastName', 'title', 'role', 'role_other', 'institution', 'city', 'state', 'country', 'latitude', 'longitude', 'has_image']
+    fields = ['id', 'firstName', 'lastName', 'title', 'role', 'role_other', 'institution', 'city', 'state', 'country', 'latitude', 'longitude']
     response = JsonResponse({'data': list(query.values(*fields))})
     response['Access-Control-Allow-Origin'] = '*'
     return response
--- a/django/users/views.py
+++ b/django/users/views.py
@@ -36,15 +36,15 @@ def retrieve(request):
     # get mailing list status (mailchimp uses md5 hashes of emails to identify subscribers)
     email = data["email"].lower().encode(encoding='utf-8')
     subscriber_hash = hashlib.md5(email).hexdigest()
-    mc_client = MailChimp(mc_api=settings.MAILCHIMP_KEY, timeout=10.0)
 
     try:
+        mc_client = MailChimp(mc_api=settings.MAILCHIMP_KEY, timeout=10.0)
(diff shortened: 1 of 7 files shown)


django/users/views.py
```diff
@@ -126,22 +128,22 @@ def register(request):
...
-
-        d = {'firstname': created_user.firstName, 'url': url}
-
-        subject, from_email, to = 'BRCAExchange account confirmation', 'noreply@brcaexchange.org', created_user.email
-        text_content = plaintext_email.render(d)
-        html_content = html_email.render(d)
-        msg = EmailMultiAlternatives(subject, text_content, from_email, [to])
-        msg.attach_alternative(html_content, "text/html")
-        msg.send()
+        # # Send activation email
```
melissacline: Why is this commented out rather than removed?  If your thinking is to keep it accessible in case we want to add it later, then it's not worth commenting out because (a) it's in the github history and (b) when/if we want to add it later, the details might not be the same at that time, so we'd be writing new code anyway.

website/js/index.js
```diff
@@ -57,7 +57,7 @@ import { Container as Grid, Col, Row, Table, Button, Modal, Card, Collapse } fro
 import {VariantTable, ResearchVariantTable, researchModeColumns, columns, researchModeGroups, expertModeGroups} from './VariantTable';
 import Signup from './Signup';
 import {Signin, ResetPassword} from './Signin';
-import {ConfirmEmail} from './ConfirmEmail';
+// import {ConfirmEmail} from './ConfirmEmail';
```
melissacline: Here's my same comment as earlier: why is this commented out rather than edited out?

website/js/Signin.js
```diff
@@ -162,33 +162,35 @@ class SigninInner extends React.Component {
...
-                                }
-                            });
-                        };
-                        var activationMessage = (
-                            <span>
-                                This account has not yet been activated. Please check your email for an activation link, or <a href="#" onClick={resendActivation}>resend activation</a>.
-                            </span>);
-                        this.setState({error: activationMessage});
-                    } else if (error.non_field_errors === 'Unable to login with provided credentials.') {
+		    // Email verification step disabled - Uncomment to renable
```
melissacline: Same comment as before: why is this commented out rather than edited out?  The commented-out code, unfortunately, makes the (real) code a bit harder to parse.

melissacline (changes requested):
I'm excited to see this moving!  But, I'm concerned that it looks like a lot of code has been commented out but not removed.  I'm concerned because in such cases, the commented-out code might not help us in the future (when/if we need it, their interface probably will have changed anyway), but leaves the codebase a bit larger and bulkier.  And if we want to refer back to the old code, we've got it in github history.

GeoffLyle:
@melissacline I believe I commented some features out since I didn't know at the time whether we wanted to fully remove them, or would want to reimplement them after review.  It looks like we are happy with the changes so I will go ahead and remove those sections that I commented out.

django/users/views.py
```diff
@@ -316,21 +296,14 @@ def users(request):
...
-    whitelist = ['id','email','firstName','lastName','title','role','role_other','institution','city','state','country','phone_number','hide_number','hide_email','is_active','is_admin','has_image','is_approved']
+    whitelist = ['id','firstName','lastName','title','role','role_other','institution','city','state','country','is_active','is_admin','is_approved']
 
     count = query.count()
 
     start = page_num * page_size
     end = start + page_size
     data = list(query[start:end].values(*whitelist))
 
-
```
e-t-k, line 327: Note to self: this deletion is OK because `email` and `phone_number` are removed from the whitelist above.

website/js/Signup.js
```diff
@@ -308,45 +308,21 @@ class SignupForm extends React.Component {
...
-
     render() {
         var onChange = function() {
             var value = this._refs.role ? this._refs.role.value : "NONE";
             this.setState({otherRole: Role.other(value)});
         };
         return (
             <div className="form-horizontal" onChange={onChange.bind(this)}>
-                {this.renderImageUpload('image', 'Profile picture')}
                 {this.renderTextInput('email', 'Email')}
```
e-t-k, line 326: Can we add a note for the user that email addresses are (now) private? eg

```suggestion
                {this.renderTextInput('email', 'Email (will not be displayed)')}
```

e-t-k (approved):
Note that I haven't looked at how this renders, just the code changes -- but looks good in general. I agree with Melissa that it's preferable to delete code rather than comment it out.

